PR #545, opened 2014-12-25
Thermometer chart

prijip:
I was in need of a Thermometer Chart for one of the reports I was working on, so I went ahead and added that capability.  This is based on JFreeChart's ThermometerPlot.

Please see below the screenshot of the updated "Edit Chart" window
chartedit

Please see below the screenshot of a sample Thermometer Chart
sampleplot


tmorgner:
Thank you! I cherry picked the three relevant commits and additionally converted the string of thermometer units into a enum. This eliminates the manual mapping from the code and allows us to reuse the existing EnumPropertyEditor instead of handcoding the mapping in a custom editor.

tmorgner:
As Github does not like manual, selective merges, I have to close this pull request without marking it merged. However, the feature is merged and available in Master via Pull Request 548.

PR #550, opened 2015-01-09
PRD-4763 - Using PRD 5.0.0.1-GA cannot open reports from 4.8.X repositor...

kolinus:
...y

What was done:
1) removed UI widgets.
2) changed getVersion method to always return 5


tmorgner:
Looks very good. Merged.

PR #552, opened 2015-01-12
[PRD-5343] - Libpensol now refreshes local tree cache when a new File is created in a JCR repository

graimundo:
- Libpensol now refreshes local tree cache when a new File is created in a JCR repository
- Added WRITE_CONTENT capability to the libpensol PentahoSolutionFileProvider


graimundo:
@tmorgner please review

PR #556, opened 2015-01-14
[PRD-5286] fixed the issue and refactored the saving actions

andrey-markelov:
@tmorgner


tmorgner:
Apart from the refactoring that extracts a AbstractSaveAction class, I cannot find any actual code change. The refactored code does exactly the same as before. 

So please help me understand what is the cause of the bug and where is the fix located in this code?

andrey-markelov:
Hi Thomas. Problem was in line 110 of SaveReportAsAction.java where code
```report.setBundle(bundle);``` replaced the "Writable" document bundle with the "Static" document bundle. I have removed this string by analogy with SaveReportAction. Also I have encapsulated generic code of saving logic in abstract class.

tmorgner:
Wow, the silly code did that? Nice find! 

One last request: Can you please write a Unit test for this fix in the "report-designer" project? Fixes without unit-tests run the real risk of regressions a year or two down the road.

dkincade:
I am processing this pull request now - please add the unit tests in a separate pull request...

PR #566, opened 2015-01-27
PRD-5184 - Metadata for CDA data-factory is wrongfuly claiming experimen...

tmorgner:
...tal status for CDA datasource, but community supported is correct. Open-ERP no longer needs to specify deprecated "experimental" flag when maturity level is set.

(cherry picked from commit a293446)


tmorgner:
Damn, tagged as wrong case number in the commit. 

@dkincade please give it a merge. Due to the rollback of changes in the UI to avoid the cost of doc updates, we now only filter by "community supported" for the 5.x branch. 6.0 with the market-place shall see more levels. 

In the meantime, that means that the CDA datasource should have been promoted to "Community supported" instead of experimental. After nearly 3 years of service with no bugs reported it is safe to call it "safe to use".

PR #567, opened 2015-01-28
[BACKLOG-2035] In PRD, OK button on MongoDB data source dialog never gets enabled.

mdamour1976:


mdamour1976:
@dkincade @tmorgner for review/merging please

PR #569, opened 2015-01-29
[PRD-5353] - Regression: Hide Parameter UI System Parameter on Drilldown...

aliakseihaidukou:
... function downloads a PRPT report instead of hiding the parameter UI of the target report

switched from "content" to "viewer?showParameters=false" URL

@tmorgner please review


tmorgner:
Yes, it is on the wiki. I don't blame you that you didn't find it there, as the Wiki is no longer the standard place for documentation.

http://wiki.pentaho.com/display/Reporting/URL+Parameters+for+the+Report+Viewer+Plugin

PR #571, opened 2015-02-02
PRD-5184  - Ever since its inception, the parameter editor never filtere...

tmorgner:
...d data-sources other than by visibility. This lack of filtering goes back to at least version 3.7.


tmorgner:
@dkincade More on the same.

PR #572, opened 2015-02-04
[PRD-5354] [dialog-validate] Able to create parameter without name, but ...

ivanpogodin:
...not able to delete it and save report

@tmorgner review it please

NOTE: package org.pentaho.reporting.designer.core.settings.ui is used for validation
TODO: localization


ivanpogodin:
@tmorgner could you review it?
The solution is
  User is unable to create parameter without name. Alert appears instead and the param-dialog stays open.
note
Alert text is i18n property. There is a _ja localization in the project. I'm not good at Japaneze.
Anybody else is required to translate it.

ivanpogodin:
@tmorgner
see also https://github.com/pentaho/pentaho-reporting/pull/584 (SP-1684 - backport to 5.3)

PR #584, opened 2015-02-25
[SP-1684] Backport of PRD-5354 - Able to create parameter without name, ...

ivanpogodin:
...but not able to delete it and save report (5.3 Suite)

@tmorgner could you review it?
see also https://github.com/pentaho/pentaho-reporting/pull/572 (PRD-5354)


tmorgner:
@dkincade @mdamour1976 I tested the master build and it is fine. The patch here is identical to master. It might be worth setting up the 5.3 builds in CI so that builds can pull in 5.3 branch artefacts when we do tests of the SP commits.

PR #591, opened 2015-03-16
PRD-5393 - Regression: Report breaks after setting property widow-orphan...

Aliaksandr-Kastenka:
...-opt-out to false on the Group


Aliaksandr-Kastenka:
@mdamour1976 
Please review.

PR #593, opened 2015-03-18
[PRD-5340] - Item label on Line chart is cut off at the top - trying to ...

akhayrutdinov:
...enlarge axis range to let item label be pictured inside chart's area

@tmorgner review it please.

### DIFF
--- a/engine/legacy-charts/source/org/pentaho/plugin/jfreereport/reportcharts/AreaChartExpressionBundle.properties
+++ b/engine/legacy-charts/source/org/pentaho/plugin/jfreereport/reportcharts/AreaChartExpressionBundle.properties
@@ -335,6 +335,14 @@ property.rangeMaximum.description=
 property.rangeMaximum.deprecated=
 
 
+property.scaleFactor.display-name=y-scale-factor
+property.scaleFactor.grouping=Y-Axis
+property.scaleFactor.grouping.ordinal=600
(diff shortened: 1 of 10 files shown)


tmorgner:
This patch will cause troubles. Your code completely disables the auto-range feature, which will break reports when customers rely on that setting. 

Please restore the auto-range feature. Then make the scale-factor a property on the chart expression that can be edited by the user, and default it to 1.0 (this makes sure that existing charts do NOT change - nothing upsets customers more than reports that suddenly look different).

Second, ensure that it is documented that the padding does not work with the auto-range feature. Alternative solution for auto-range = true (depending on the customer wishes and PM scope), implement the auto-range feature by iterating through the chart data-set locating the minimum and maximum value and adding padding to them and then set the upper and lower range bounds accordingly.

But as it is, this fix should NOT be merged.

akhayrutdinov:
@tmorgner, I sent you a letter. Take a look at https://github.com/akhayrutdinov/pentaho-reporting/commit/221c293ad9e14f6a4b2e7659e54125160f281a26 please. Just to make sure I've understood you clearly and not to send a meaningless pull-request :)

akhayrutdinov:
@tmorgner,

this is an updated version. I also have cut comparison with 1e-4 and replaced it with a mere greater than operator. Review it please

tmorgner:
Looks great. Merged.

PR #596, opened 2015-03-22
Added query hash and referenced fields support to CDA queries to enable data/metadata caching

ITXpander:
Fixes: http://jira.pentaho.com/browse/PRD-5414


tmorgner:
Thank you!

PR #597, opened 2015-03-30
[PRD-5332] - wrong java version taken in Windows 7 with multiple java ve...

akhayrutdinov:
...rsions installed

Changed the bat so, that it will respect _PENTAHO_JAVA_HOME variable; also updated the comments that relate to JDK's version

@tmorgner, @mkambol, @mdamour1976 review it please


tmorgner:
Looks sane. Merged.

PR #606, opened 2015-03-31
PRD-5384 - Get publish success message when user has no write permission...

Aliaksandr-Kastenka:
... to target location

Goes with https://github.com/pentaho/pentaho-platform/pull/2270


Aliaksandr-Kastenka:
@tmorgner
Please review.

tmorgner:
In conjunction with that platform change, this looks great!

PR #610, opened 2015-04-17
[ENGOPS-640] Added javadoc package path

abaskakau:
Added javadoc.packagenames property for dist-javadoc target
@smaring @lgrill-pentaho


abaskakau:
@mtseu FYI

PR #617, opened 2015-05-06
5.2

wsfyy:


tmorgner:
We do not merge old branches back into master.

PR #618, opened 2015-05-18
BACKLOG-3306 - Checkin 5.4 Localization files from Acclaro (2nd drop)

kolinus:
What was done:
1) updated the files received from the company


kolinus:
@tmorgner @mkambol  could you please review

mkambol:
@kolinus Your comment says that _en files were added, but I only see _ja, _fr, and _de.  Were any files missed in this commit?

kolinus:
@mkambol  thank you for double checking -- just forgot to remove the line in comment when copy pasted it from similar pull requests.
No en files are missed -- they are not necessary for reporting.

PR #625, opened 2015-06-02
PIR-1142 - PIR: Adding two fields to groups the font of the second on…

Aliaksandr-Kastenka:
…e is the same color as the background
Goes with https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/pull/234


Aliaksandr-Kastenka:
@tmorgner 
Please review.

PR #630, opened 2015-06-11
[PRD-5456] - Element's text color attribute does not work when report…

akhayrutdinov:
… is exported to MS Excel

 - do not cache Excel cell style for elements with different style sheets
 - add a test for caching producer

@tmorgner, review it please


akhayrutdinov:
@tmorgner, review it please

PR #632, opened 2015-06-15
[BACKLOG-3680] - Upgrade hypersonic library on pentaho-reporting and …

aliakseihaidukou:
…its sub modules

ivy.xml - hsqldb switched from 1.8.0 to 2.3.2
sampledata.script is updated, it is identical in all places.
Prd-5321 - changed sql query - added "order by" clause - default ordering changed, tests were failing. Now order will be the same if smth changes.
txt queries - changed with the driver change.
steelwheels.mondrian.xml - now uses DIM_TIME instead of TIME to use common sampledata script.
Prd-4343.prpt - changed "where" parameters - TIMESTAMP requires time now("YYYY-MM-DD HH:MM:SS")

@rmansoor @tkafalas please review


tmorgner:
Looks good and sane. Thank you for summarising the changes - this does help a lot when reviewing the code.

PR #636, opened 2015-06-23
[SP-1912] Backport of PRD-5456 - Element's text color attribute does …

ivanpogodin:
…not work when report is exported to MS Excel (5.4 Suite)
@tmorgner, review it please


tmorgner:
@rmansoor This is good, please merge. Straightforward port from master code.

PR #637, opened 2015-07-13
Fix a wrong value of minutes that sometimes are greater than expected…

serval2412:
… value.

See https://bugs.documentfoundation.org/show_bug.cgi?id=92693


dkincade:
Please provide a unit test which illustrates this defect before it is fixed and then passes with the fix being in place.

serval2412:
I don't know how to do this, let's forget it.

PR #639, opened 2015-07-15
[CLEANUP] Applying Pentaho Checkstyle to Pentaho Reporting Libraries

dkincade:


dkincade:
Validated that all the unit tests in the library section contiue to function after the reformat

PR #644, opened 2015-07-27
[SP-2008] - Backport of PRD-5363 - Sub-report: AskTimeOutException ap…

akhayrutdinov:
…pears when user opening a sub-report. (5.4 Suite)

- recreate the actor if it was stopped

@tmorgner, review it please


tmorgner:
Looks good. restarting the dead actor is the right thing to do, and no harm can come out of it.

PR #650, opened 2015-08-14
[CLEANUP] - Fixes for some broken tests

tmorgner:
The Prd5321 test cannot be run safely on a CI machine as it uses local font rendering, and thus requires careful setup. Tests like that probably need to run in a carefully curated test environment, like a VM or so.

GoldenSample tests were failing after an overly eager formatting round on the various XML files changed some property values. Adding new-lines randomly is bound to cause troubles.

Finally: Updated the IntelliJ module files to include the mockito jars used by LibFormula and LibFonts on their classpath.

### DIFF
--- a/engine/core/test/org/pentaho/reporting/engine/classic/core/bugs/Prd5321Test.java
+++ b/engine/core/test/org/pentaho/reporting/engine/classic/core/bugs/Prd5321Test.java
@@ -215,6 +215,10 @@ public class Prd5321Test {
 
   @Test
   public void testTextRenderingComplex() throws Exception {
+    if (DebugReportRunner.isRunFromAnt()) {
+      return;
+    }
+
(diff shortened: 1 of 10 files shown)


tmorgner:
@dkincade Please review, merge and hope that CI is not blasting out its anger any more.

dkincade:
I am still getting a failure on TableLayoutTest on testFixedSizeTableCellsRelativeSizeCanvasComplex when I run from ant locally with these changes. Any advice?

tmorgner:
Added another exception. Given that I work on Windows, where all standard fonts are available, I cannot verify which tests will fail until it build on a barebones Mac or Linux system.

If you see Complex in the test name, it most likely refers to the complex text layout - and thus will fail.

dkincade:
I am still getting failed tests locally so I expect this to still fail in CI and have to disable more tests...

PR #651, opened 2015-08-14
PRD

tmudarikwa:


tmorgner:
No, we do not merge bug-fix branches into master. When a bug is fixed in the old branch, it gets also committed to the master, either via cherry picking or manual merging at that time.

PR #652, opened 2015-08-15
[CLEANUP] - More test fixes

tmorgner:
CI did not pick up my exclusion, so now I use the conservative measure of demanding a specific system property to be set before the dangerous tests are run. Ant, or any normal developer, will not have the system property set, so the tests are skipped. 

Kettle tests were failing as we did not pull in VFS2 and Guava via Ivy. The tests are much happier now.


tmorgner:
@dkincade More fixes, that should shut up the engine parts now.

PR #655, opened 2015-08-24
[BACKLOG-4883] Get Report Designer.app working without java mac compa…

pedrofvteixeira:
…t issues

	- added 2 features to 'bundleapp' target ( OSX's app builder ):
		- include system/karaf structure within .app's root dir
		- set workdir to be the .app's root dir via '-Duser.dir' option ( used in KarafBoot )

### DIFF
--- a/designer/report-designer-assembly/build.assembly.sequential.xml
+++ b/designer/report-designer-assembly/build.assembly.sequential.xml
@@ -249,8 +249,18 @@
     </copy>
   </target>
 
+  <target name="stage-mac.copy-pentaho-system">
+    <!-- leverages on existing 'resolve-pentaho-system' target --> 
+    <antcall target="resolve-pentaho-system" inheritall="true" inheritrefs="true" />
+    <!-- place pentaho-system resources on a system folder, at the root of the .app -->
+    <mkdir dir="${report-designer.dir.mac}/system"/>
+    <copy todir="${report-designer.dir.mac}/system" overwrite="true">
+      <fileset dir="${system.dir}"/>
+    </copy>
+  </target>
+
   <target name="stage-mac.finalize-bundle"
-          depends="stage-mac.inject-filetypes, stage-mac.copy-files, stage-plugins-mac"/>
+          depends="stage-mac.inject-filetypes, stage-mac.copy-pentaho-system, stage-mac.copy-files, stage-plugins-mac"/>
 
   <target name="stage-mac" depends="init, install-mac-app-bundler" if="mac.java.runtime.home">
     <echo>Using JDK from ${mac.java.runtime.home}</echo>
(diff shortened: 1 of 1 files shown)


pedrofvteixeira:
@rmansoor , @CodeOnCoffee , @mbatchelor please review

pedrofvteixeira:
@dkincade ^

tmorgner:
The commit reroutes the "user.dir" system property. We use this property to locate configuration files within the user's home directory. A side effect of that will be that the configuration files for both Pentaho Reporting and Kettle will be stored inside the app-bundle, not in the user's home directory. 

This will also affect how kettle locates plugins, and thus it will not see plugins installed in the "$HOME/.kettle" directory.

Note that defining this property is not the same as setting the working directory of an application. "new File(".").absoluteDir().toString()" will still tell you that your current working directory is $HOME.

What is the reason that the "user.dir" property has to be redefined?

pedrofvteixeira:
Hi Thomas,

There are 2 system properties in play: "user.home" and "user.dir"

As far as which one gets used, pentaho-reporting uses *the first one* ( "user.home" ) as the $HOME dir definition:

	- In SmarterContextFactory, for the ".pentaho/simple-jndi" base dir : https://github.com/pentaho/pentaho-reporting/blob/5.4.0.3-R/designer/report-designer/src/org/pentaho/reporting/designer/core/util/jndi/SmarterContextFactory.java#L42
	
	- In FileConfigStoreModuleInitializer, for the storage basePath: https://github.com/pentaho/pentaho-reporting/blob/5.4.0.3-R/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/misc/configstore/filesystem/FileConfigStoreModuleInitializer.java#L104
	
	- In BinaryPreferencesFactory, for ".pentaho/report-designer" base dir: https://github.com/pentaho/pentaho-reporting/blob/5.4.0.3-R/designer/report-designer/src/org/pentaho/reporting/designer/core/settings/prefs/BinaryPreferencesFactory.java#L40
	
	- in configuration.properties, as the file-data base path: https://github.com/pentaho/pentaho-reporting/blob/5.4.0.3-R/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/misc/connections/configuration.properties#L4



I have not found explicit calls to "user.dir" in pentaho-reporting, that is why the change has been applied to this system.property
 
	- setting "user-dir" path *does* change the working.dir:

		- "new File('.').getAbsolutePath()" without setting this property returns '~/'

		- "new File('.').getAbsolutePath()" when setting this property to $APP_ROOT returns the proper path to the .app's base dir: http://pasteboard.co/9cRKbOR.png 



Nevertheless, and given these concerns, I can propose an alternative solution for your analysis:

	1 - in the bundleapp ant task, set a new system property ( something like 'osx.app.work.dir' ) with the $APP_ROOT value ( instead of setting "user.dir" )

	2 - in platform-extensions' KarafBoot, whenever a '/system/karaf' *is not found* in the standard/default dir *and* such a system property exists, then use that one.


Your thoughts on this ?

Also, @rmansoor , @CodeOnCoffee , @mbatchelor

rmansoor:
@pedrofvteixeira I am ok with your alternative approach.

pedrofvteixeira:
Closing this PR in favour of the alternative sugested approach

PR #658, opened 2015-08-27
PRD-5522 Enable excel indent

flatrixnet:
See related issue.

Fully working implementation for excel 2007 indent feature


tmorgner:
Just one change: At the moment your test dumps the file into the root directory, where it shows up in Git has untracked file. 

Can you move the output into the "test-output" directory? For that, simply call "org.pentaho.reporting.engine.classic.core.testsupport.DebugReportRunner#createTestOutputFile()" and then change your filename to "test-output/Prd-5522.xlsx". 

The "test-output" directory is covered by a gitignore rule.

flatrixnet:
Changed as requested

tmorgner:
There are multiple .gitignore files. The one in the root covers all the common stuff, and the one in "engine/core" covers the test-specific excludes.

flatrixnet:
Ok, Perfect

PR #663, opened 2015-09-02
[BACKLOG-5037] Updated the build process to add PermGen space setting…

dkincade:
…s to Info.plist


dkincade:
@pentaho-nbaker @rmansoor @lgrill-pentaho please review ... we want to get this into the next build and then we can address the structure of the Pentaho Report Designer on OS X in the next few days.

PR #675, opened 2015-09-22
[PRD-2736] - For large text strings (without white space, i.e URLs) word wrap is not enabled within an element

akhayrutdinov:
- introduce a brand new interface (SplittableRenderNode) for those nodes, that can split themselves horizontally
- make RenderableText implement the interface
  - the split is done only if WORDBREAK property is set to true
- change AbstractAlignmentProcessor to handle words splitting
  - do the separation only in there are no other possibilities to cram the word inside desired bounds
  - do nothing when proceeding the last line
- add tests
- change legacy tests so, that they will not fail because of splitting routine
  - comment each change pointing out its reason

@tmorgner, review it please.
Per @pamval's hint separation is done only if ```WORDBREAK``` property is ```true```.


akhayrutdinov:
@tmorgner, review it please

tmorgner:
This code looks very good. Let me merge it ..

PR #678, opened 2015-10-02
[BACKLOG-5400] Split UT/IT for project:pentaho-reporting-engine-class…

annapopova1:
…ic-core

This project contains 693 tests.
In my opinion the most of this tests are integration tests because they check whole report object but not concrete small piece such as concrete method of class. Furthermore it uses real prpt/xml files for testing...
All such tests I renamed as integration.

In addition I add "inttestsrc.dir=${basedir}/test" property to build.properties file for executing integration tests.

@dkincade @tmorgner Could you review my PR and let me know if I should use another criteria for separating tests?


dkincade:
@mdamour1976 @rfellows ... take a look (I will not be able to do a decent look at this for a while)

rfellows:
Looks good. Almost all changes are a direct result of class renames. the few that are not are splitting out portions of a test into an IT and leaving the real unit test(s) alone

PR #680, opened 2015-10-11
5.4.0.1

murraybower:


tmorgner:
We do not blindly merge master into a bug-fix only branch. Closed.

tmorgner:
If this is a valid pull request, please reference a case number with it or at least a description on why this change should be done..

PR #682, opened 2015-10-21
[BISERVER-11565] - File upload names fail if the name has double quotes

akhayrutdinov:
- use new endpoint, allowing encoded names
- add integration tests
- add jersey-test-framework as test dependency

@tmorgner, review it please.
This PR uses the endpoint which was introduced by https://github.com/pentaho/pentaho-platform/pull/2655

I added ```jersey-test-framework``` as test dependency (copied from ```platform```).


tmorgner:
Looks good. Merged.

PR #684, opened 2015-10-22
[BACKLOG-5473] - Rename tests' folders in pentaho-reporting to "test-src"

akhayrutdinov:
@tmorgner, review it please.
In ```engine/wizard-core```, I ignored one test. It had been failing before I made the changes, but since the folder's name didn't follow the common pattern CI knew nothing about it.


tmorgner:
Sweet! Merged! Would you mind creating a JIRA case for the failing wizard test? Either it indicates a bug that needs fixing or a test that is obsolete - in either case it needs a code change to fix.

PR #687, opened 2015-10-23
PRD-5561 - Copy & Paste error in CSVARRAY formula function

tmorgner:

PR #695, opened 2015-11-17
[SP-2218] - Backport of PRD-5561 - Libformula Copy/Paste bug in CsvArrayFunction (6.0 Suite)

akhayrutdinov:
- backport 7d940d16914a692378884bb6841241c2943d3d2f

@tmorgner, review it please. This is a backport of https://github.com/pentaho/pentaho-reporting/pull/687

Also, please look at my comment there. You test passes always -- no matter was the fix applied or not


tmorgner:
Yeah, it would have helped if the last line if I had not left that last line of the test commented out. :)
Merged!

PR #701, opened 2015-11-19
[BACKLOG-5384] Fixed failed unit tests

annapopova1:
The reason of failing was junit assertions were before setting right values. The TextComponentEditHandler uses SwingUtilities.invokeLater method. I also used the same method for assertions. Another way would be using Thread.sleep... but I prefer SwingUtilities.invokeLater.
@tmorgner could you review?


tmorgner:
This wont actually work as you'd expect. JUnit waits for thrown exceptions on its own thread. So when your code runs on the Event-Dispatcher Thread (or any other thread other than the main JUnit thread), this exception will throw, and no one will react to it. 

If you do asyncronous testing, your asynchronous code will have to catch these errors (catch Throwable, which catches both exceptions and errors), store it somewhere safe. Your main JUnit thread then has to block until for the async call finishes, and then you can check whether there is an exception stored somewhere in your thread and potentially rethrow it on the JUnit thread.

annapopova1:
@tmorgner could you take a look one more time?
I used CountDownLatch to make sure that thread was finished and checked assert concrete background color.

tmorgner:
Thank you. That looks safe now.

PR #704, opened 2015-11-25
[PDI-14499] - Randomly missing header/footer in PDF created by Pentaho Reporting file output step

akhayrutdinov:
- amend rollback state selection - if the last visible state equals current restore state, then it is necessary to rollback to the page start
- add case-specific integration test and a utility class

@tmorgner, review it please


akhayrutdinov:
@tmorgner, review it please

tmorgner:
Merged this request via an indirect merge on #706 . To validate this patch I had to regenerate the golden samples, as they were broken since [PRD-2736] was implemented. That case added line-breaking for long words, which caused changes to the computed layout of the samples.

This patch here also causes changes. It seems that some reports suffered from invisible flaws related to this case. As a result of the more accurate roll-backs, some box sizes changed as break-markers are now inserted at the correct spot.

PR #712, opened 2015-12-11
[PRD-5556] - Running PRPT in the background increases memory usage

akhayrutdinov:
@tmorgner, review it please. Here are commits' messages:
 
[CLEANUP] - Add test-lib to LibPensol as Mockito is needed here. Update IntelliJ Code-Style rules to match Pentaho code style.
(cherry picked from commit 7e49ac8)

[PRD-5556] - Running PRPT in the background increases memory usage
- store derived instances; this severs the link to the master-report and marginally reduces retained heap size (cherry picked from commit c5d6d2f)
- introduced general process-level data store; this allows us to easily share instances across functions and helps to reduce heap usage (adopted from commit fdca215)
- optimize style resolver (adopted from commit a921b54)
  - by sharing more caches and instances, we can greatly reduce the retained heap here;
  - each instance used to use 4KB of heap, now with sharing we are down to roughly 250 bytes
- tone down the logging a bit (cherry picked from commit 74fdc0b)
- if a map is empty, use global shared empty instance (cherry picked from commit 0e80511)
  - potential bug: do not mix attributes and parameter!
- reorganize internal data for attribute-map (adopted from commit fbfc238)
  - using a single map saves quite some memory without too much negative impact
  - having a single key-set allows for some faster iteration.
- stateless classes can safely be singletons to reduce memory pressures (adopted from commit 54d83e4)
- normalize all data-attributes (adopted from commit af77380)
  - this way, if we see the same data over and over again, we can consolidate them into single instances
  - everyone else will not be worse off
- activate the fast-export scan only when we are actually in fast export mode (adopted from commit d4fd0c5)
  - every bit of data not stored on elements will reduce size manyfold as elements get cloned a lot
- wrapping style-keys into a read-only list allows safe sharing without creating copies all the time (adopted from commit d2ae4ca)
  - the key-list is only used in non-performance critical code, so it is relatively safe to do this

[PRD-5556] - Running PRPT in the background increases memory usage
- add several more improvements
  - in DataCacheKey, handle singletonMap case
  - in ParagraphPoolboxStyleSheet, optimise for-loop
  - in DefaultDataAttributeCache, make cache key's fields final
  - in ReportAttributeMap
    - remove the flag
    - introduce a separate immutable subclass
    - add tests
  - in AttributeMap
    - fix some typos
    - replace LinkedHashSet with ArrayList in getNames()
    - in getAttributes(), return empty map instead of null as the javadoc claims

[CLEANUP] - Apply Pentaho formatting
- apply proper formatting for all classes affected within [PRD-5556]

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -21,6 +21,7 @@
         <option name="GENERATE_FINAL_LOCALS" value="true" />
         <option name="GENERATE_FINAL_PARAMETERS" value="true" />
         <option name="INSERT_OVERRIDE_ANNOTATION" value="false" />
+        <option name="LAYOUT_STATIC_IMPORTS_SEPARATELY" value="false" />
         <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="666" />
         <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="666" />
         <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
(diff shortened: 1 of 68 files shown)


akhayrutdinov:
added more commits from Thomas

PR #714, opened 2015-12-17
BACKLOG-5887 - Dashboard cannot resolve parameter with a default value.

Aliaksandr-Kastenka:
@pamval @tmorgner 
Please review.

Dashboard regression took place beacuse of this issue pull-requests:
http://jira.pentaho.com/browse/PDB-1869
https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/180
https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/177

I have sent pr to revert them.
Proposed fix covers up both PDB-1869 and BACKLOG-5887.

### DIFF
--- a/engine/core/source/org/pentaho/reporting/engine/classic/core/parameters/DefaultReportParameterValidator.java
+++ b/engine/core/source/org/pentaho/reporting/engine/classic/core/parameters/DefaultReportParameterValidator.java
@@ -146,6 +146,11 @@ public class DefaultReportParameterValidator implements ReportParameterValidator
     if ( untrustedValue == null ) {
       // compute the default value
       defaultValue = parameterDefinitionEntry.getDefaultValue( trustedParameterContext );
+      if ( parameterDefinitionEntry instanceof ListParameter == true ) {
+        if ( ( (ListParameter) parameterDefinitionEntry ).isAllowMultiSelection() && defaultValue instanceof String ) {
+          defaultValue = computeValidArrayValue( defaultValue );
+        }
+      }
       untrustedValue = defaultValue;
     }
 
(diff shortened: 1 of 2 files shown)


pamval:
I processed the reporting plugin one. @tmorgner can you please take a look at this one ?

thanks

tmorgner:
This code as it is is dangerous. You cannot simply split strings on a comma. Any real value that legally contains a comma would be broken into invalid chunks. Imagine a list of Kings, with the default value of 

* Joe, the bard

With your method, you;d get two values: "Joe" and "the bard", which is incorrect. 

And of course, apart from the dangerous effect itself: 

This code is not addressing the problem at the correct level. If a multi-select prompt returns a non-array value as default value, then this in itself is an error. Find out why it is wrong.

Please check that PIR actually creates the prompt correctly. For a multi-select parameter, the prompts data type should be String array (check that in the PRPT file itself (datadefinition.xml contains the prompts). 

I attached a sample report defined in PRD that uses a multi-select string parameter. And here is some Unit-test code to demonstrate that the prompt returns an array.

    @Test
    public void testParameterDefaultValue() throws ResourceException, ReportProcessingException {
      MasterReport report = DebugReportRunner.parseLocalReport( "/PDB-1869.prpt", getClass() );
      ParameterDefinitionEntry parameterDefinition =   report.getParameterDefinition().getParameterDefinition( 0 );
      Object defaultValue = parameterDefinition.getDefaultValue( new DefaultParameterContext( report ) );
      Assert.assertArrayEquals( (Object[]) defaultValue, new String[] {"C,D"} );
    } 

The correct fix would address PIR and would make sure that a multi-select prompt has the correct data type. 

If - and only after asserting that PIR cannot be fixed - such a patch needs to go in, then at least use the CSVTokenizer class from LibBase for the job. This classs handles quoted CSV strings and at least gives you a chance of producing results that dont break the validation.

PR #716, opened 2015-12-24
PRD-5567 Tab of report displays not correctly information about stateof report.

prakapenka:
Avoid to change last file name for report in case report attempted to be save with incorrect path.
We can't determine using file.getCanonicalPath() is path is correct since file can be saved in filesystem as far as in any independent destination and only call to SaveReportUtilities.saveReport(...) will give the real picture can't this file for any reason or been saved successful.
Refactoring to use 'ReportDesignerBoot.LAST_FILENAME' in all places where 'report-save-path' string was directly hard-coded.


tmorgner:
Looks good. Merged.

PR #717, opened 2016-01-04
PRD-5511 - PRD: Not saved parameter mapping when it uses formulas

prakapenka:
The root cause of formulas not being saved is incorrect conversion of formulas parameters. Formulas since that never have been saved to report correctly. This issue being hided by arguments - so when arguments is existed for kettle query even formulas was not recognized - report been saved
correctly. If there was no arguments - formulas for kettle query were existed only in master report and never being saved to xml/zip file.

To fix that libformula/FormulaUtil.collectReferences(...) should start recognize LValue as FormulaFunction. So that we can be able to extract parameter references from formula, then - since we recognize references we will be able to create FormulaParameter in FormulaParameter.convert(
ParameterMapping[] ...) and since formulas being converted they will be processed in

FileTransformationProducerWriteHandler (line 61)

where statement definedArgumentNames.length == 0 && parameterMappings.length == 0
will not be true since parameterMappings.length will be > 0 and formulas will be saved to zip report file.


tmorgner:
Looks good, and with unit tests it is a beautiful thing. Merged.

PR #720, opened 2016-01-13
PRD-5573 Validate space reserved for "table-header'

EgorZhuk:
Prevent infinite loop with error when table 
header is higher than allowed page height.


EgorZhuk:
Will be replaced.

PR #725, opened 2016-01-26
Prd 5595 - memory leak in a flow mode

EgorZhuk:
Boxes were not cleaned up because check for the visibility was failing.
http://jira.pentaho.com/browse/PRD-5595


tmorgner:
Ahh, basic maths. I knew it :)

Thank you for fixing this, and extra thank you for the refactoring to remove the duplicate code as well

dkincade:
Thomas never claimed to be a genius at basic math - I am pretty sure I heard him say something to the opposite.

tmorgner:
Right now the test is not working, as the Prd-5595 file is missing. Once there it is added Prd-5595.prpt file added I can merge it.

EgorZhuk:
@tmorgner Sorry, my fault. Now it is present.

PR #726, opened 2016-01-27
[BACKLOG-5967] PRD Warning message log4j.xml file not exist

pedrofvteixeira:
- [PDI-14540] applied same solution that was applied to Spoon
	- new  /system/osgi directory structure does *not* conflict with existing karaf build resource ( /system/karaf )


pedrofvteixeira:
@rmansoor please review

PR #732, opened 2016-02-12
[Backlog-6426] Added a few try-catch blocks to make Hive happy.

tmorgner:


tmorgner:
@pentaho/orlandocalisbon Please have a look, and merge as needed.

.idea/codeStyleSettings.xml
```diff
@@ -135,9 +138,8 @@
           <option name="PARENT_SETTINGS_INSTALLED" value="true" />
           <indentOptions>
             <option name="INDENT_SIZE" value="2" />
-            <option name="CONTINUATION_INDENT_SIZE" value="2" />
+            <option name="CONTINUATION_INDENT_SIZE" value="4" />
```
jvelasques: @tmorgner this has value="2" in [pentaho-coding-standards](https://github.com/pentaho/pentaho-coding-standards/blob/master/codeStyleSettings.xml#L89)

jvelasques:
tested this change locally, the bug is no longer reproducible, ready to merge once the codeStyle issue has been adressed

.idea/codeStyleSettings.xml
tmorgner: I used that setting at "2" before, but it resulted in large reformatting of the file. So I altered the setting so that it matches the file's layout (which was a result of @dkincade initial formatting - I assume he's never wrong :) ).

dkincade: @dkincade has been wrong before :-D

This file should be the exact copy of https://github.com/pentaho/pentaho-coding-standards/blob/master/codeStyleSettings.xml ... or better yet should not be included at all in this project.

jvelasques:
closing this one as #733 was merged

PR #741, opened 2016-02-26

mr9d:
- Swing version of the self-drilldown.xul dialog.


mr9d:
@tmorgner please review.

tmorgner:
That looks very clean and good. I like it. Lets kill XUL forever! :)

PR #746, opened 2016-03-01
[PRD-5599] Utility methods with listener parameter

EgorZhuk:
We need them to set listeners.
@tmorgner


EgorZhuk:
Thank you for fast merge, Thomas.

PR #747, opened 2016-03-01
Build props changes to be able to override dependencies versions.

EgorZhuk:
@lgrill-pentaho @tmorgner


lgrill-pentaho:
This is not correct.

EgorZhuk:
@lgrill-pentaho what's wrong?

lgrill-pentaho:
You have created new version properties that are not in the global version properties master list.

This will break release and nightly builds that do all the version remapping.

There is a link to the global properties file here: http://devci.pentaho.com/job/master-snapshot/view/control/job/build-resources/ws/6.1-SNAPSHOT/version.properties/*view*/

Or in the build-resources project itself:
https://github.com/pentaho/build-resources/blob/master/version.properties

All of these artifacts are built in version-sync from the osgi bundles project. There is just one version revision for all of these: dependency.pentaho-osgi-bundles.revision

tmorgner:
Wow, I did not know about that master list. You learn something new every day. :)

From the "biserver-ce/assembly" ivy.xml it seems we need to use "dependency.bi-platform.revision" here, as the assembly uses "project.version", which would correspond to "dependency.bi-platform.revision" for us.

EgorZhuk:
Unfortunately I don't have access to both links.
What is the valid way to get it fixed?

1) Remove new props from build.properties
2) Change them in ivy.xml to what? :8ball: 

@tmorgner @lgrill-pentaho

EgorZhuk:
@abaskakau will fix it in master list

PR #759, opened 2016-03-17
Backlog 7181

EgorZhuk:
Set listeners in fast mode.
Thomas, please review.
@tmorgner


engine/core/source/org/pentaho/reporting/engine/classic/core/modules/output/table/xls/ExcelReportUtil.java
```diff
@@ -243,10 +226,65 @@ public static void createXLS( final MasterReport report, final OutputStream outp
...
+    try {
+      reportProcessor.processReport();
+    } finally {
+      if ( listener != null ) {
+        reportProcessor.removeReportProgressListener( listener );
+      }
+      reportProcessor.close();
+    }
+  }
 }
```
tmorgner: Where are those two static methods used? They are most definitely not referenced in the code of this class.

PR #760, opened 2016-03-17
[BACKLOG-6988] - log if debug only if interrupted exception thrown.

prakapenka:
Separately catch and re-throw ReportInterruptedException. Log stacktrace for debug level only. Interruption is not a error drama.

see https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/245


prakapenka:
@tmorgner  please review.

PR #765, opened 2016-03-18
[PRD-5562] [SP-2504] Allow to select queries with duplicated names

EgorZhuk:
JTree expands the first node with a given name.
I didn't managed to get it working in the right way without reordering of nodes.
So, currently you can select query with duplicated name but there is a side effect - selected query ( and  its data source ) is moved forward.
I'm not sure if it is accepatble tradeoff.

@tmorgner


akhayrutdinov:
@tmorgner, review it please

tmorgner:
This seems not a bug at all: I referred the case back to triage. 
http://jira.pentaho.com/browse/PRD-5562

EgorZhuk:
Not a bug

PR #767, opened 2016-03-21

mr9d:
- Fix for the bug with pentaho-sugar-drilldown dialog that appeared in the latest QAT build.


mr9d:
I've fixed a minor bug, could you please merge the fix, @tmorgner?

PR #769, opened 2016-03-22
[BACKLOG-7159] Add total pages field to report progress event

EgorZhuk:
Thomas, please review.
@tmorgner

### DIFF
--- a/engine/core/source/org/pentaho/reporting/engine/classic/core/layout/output/AbstractReportProcessor.java
+++ b/engine/core/source/org/pentaho/reporting/engine/classic/core/layout/output/AbstractReportProcessor.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2001 - 2013 Object Refinery Ltd, Pentaho Corporation and Contributors..  All rights reserved.
+ * Copyright (c) 2001 - 2016 Object Refinery Ltd, Pentaho Corporation and Contributors..  All rights reserved.
  */
 
 package org.pentaho.reporting.engine.classic.core.layout.output;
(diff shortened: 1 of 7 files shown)


engine/core/source/org/pentaho/reporting/engine/classic/core/layout/output/AbstractReportProcessor.java
```diff
@@ -1460,7 +1460,7 @@ public void processReport() throws ReportProcessingException {
             + ": Report processing failed.", e );
         throw new ReportProcessingException( "Failed to process the report", e );
       }
-      fireProcessingFinished( new ReportProgressEvent( this, getPhysicalPageCount() ) );
+      fireProcessingFinished( new ReportProgressEvent( this, getPhysicalPageCount(), getLogicalPageCount() ) );
       if ( AbstractReportProcessor.logger.isDebugEnabled() ) {
```
tmorgner: Your change uncovered a bug. The "getPhysicalPageCount()" call is wrong. It should be the logical-page we are querying here (to keep it in common with all the other page-event uses of the page-count). 

The reporting engine can have reports where multiple physical pages form a single logical page (page-spanning or poster-print mode). 

Whenever we query the total pages (@see calculatePageCount(..)) we actually count the logical pages (or in other words: How many page-breaks did we detect.) Thus for the sake of consistency, change this "getPhysicalPageCount" call into a getLogicalPageCount() call. Then our process-state never reports a page-number that is larger than the total page number.

tmorgner:
Can you add some unit-tests around the "reuse" methods to make sure they work as expected and that they preserve the total-page-count if they dont explicitly update it?

engine/core/source/org/pentaho/reporting/engine/classic/core/layout/output/AbstractReportProcessor.java
EgorZhuk: The same issue in plugin?
Code from CachingPageableHTMLOutput
```
private IReportContent produceReportContent( final PageableReportProcessor proc,
                                               final ZipRepository targetRepository )
    throws ContentIOException, IOException {
    //physical pages again?
    final int pageCount = proc.getPhysicalPageCount();
    final ContentLocation root = targetRepository.getRoot();
    final Map<Integer, byte[]> pages = new HashMap<>();

    for ( final ContentEntity contentEntities : root.listContents() ) {
      if ( contentEntities instanceof ContentItem ) {
        final ContentItem ci = (ContentItem) contentEntities;
        final String name = ci.getName();
        final int pageNumber = extractPageFromName( name );
        if ( pageNumber >= 0 ) {
          pages.put( pageNumber, read( ci.getInputStream() ) );
        }
      }
    }
    return new ReportContentImpl( pageCount, pages );
  }
```

@tmorgner

EgorZhuk:
ok, will fix and add tests

engine/core/source/org/pentaho/reporting/engine/classic/core/layout/output/AbstractReportProcessor.java
tmorgner: Yes. Same problem here.

EgorZhuk:
Replaced with #770

PR #773, opened 2016-04-11
[PRD-5535] - Count Distinct (Running) on field 'Field Name' doesn't d…

apratkin:
…isplay a drop-down with default JDBC fields


apratkin:
@tmorgner Could you please review my commit?

tmorgner:
This is totally fine. Merged.

PR #784, opened 2016-05-20
[BACKLOG-7998] - Performance: Reports with TotalGroupSumFunction run …

prakapenka:
replace hash-map implementation with array-based structure.


prakapenka:
@tmorgner please review

tmorgner:
This looks really good. 

Given that similar code will be needed elsewhere (TotalItemCount, TotalItemMin/max use the same Hashmap<key,sequence>-structure) - can you extract the code that is dealing with the two array-lists into a separate, reusable object? This makes reusing it trivial and also allows you to properly unit-test it (which brings up the coverage numbers and makes Wingman happy).

tmorgner:
@prakapenka Wingman still wants more coverage.

EgorZhuk:
See #786

tmorgner:
I think this PR can be closed.

PR #785, opened 2016-05-25

mr9d:
- Missed Target and Tooltip fields

### DIFF
--- a/designer/report-designer-extension-pentaho/source/org/pentaho/reporting/designer/extensions/pentaho/drilldown/swing/SwingRemoteDrillDownUi.java
+++ b/designer/report-designer-extension-pentaho/source/org/pentaho/reporting/designer/extensions/pentaho/drilldown/swing/SwingRemoteDrillDownUi.java
@@ -62,10 +62,10 @@ public class SwingRemoteDrillDownUi implements DrillDownUi {
   private JPanel editor;
 
   /** Server URL input field. */
-  JTextField serverUrlField;
+  private JTextField serverUrlField;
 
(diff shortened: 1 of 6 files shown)


designer/report-designer-extension-pentaho/source/org/pentaho/reporting/designer/extensions/pentaho/drilldown/swing/SwingRemoteDrillDownUi.java
```diff
@@ -56,6 +61,15 @@
   /** Top=level components container. */
   private JPanel editor;
 
+  /** Server URL input field. */
+  JTextField serverUrlField;
```
tmorgner: Is there a specific reason why these fields are package protected? Shouldn't they be private instead? If you needed to access them or their contents, accessors would do that job more safely.

mr9d: Nope, I just missed it. Fixed, thank you.

mr9d:
@pamval could you please rerun Wingman for this PR?

tmorgner:
@wingman-pentaho is lazy, that bugger only monitors the engine/core project, not the other modules. But then again, poor Wingman is locked away in an ice-cold server room with no sunlight, so no wonder he's not willing to look at side projects. I would neither if I were about to freeze to death :)

mr9d:
Very comprehensible explanation =)

PR #786, opened 2016-05-30
[BACKLOG-7998] - Improve perfomance of 'Total' functions

EgorZhuk:
Finishing @prakapenka work.
Junits are on the way - need to see wingman's opinion.
@tmorgner

### DIFF
--- a/engine/core/source/org/pentaho/reporting/engine/classic/core/function/StateSequence.java
+++ b/engine/core/source/org/pentaho/reporting/engine/classic/core/function/StateSequence.java
@@ -1,0 +1,102 @@
+/*
+ * This program is free software; you can redistribute it and/or modify it under the
+ *  terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ *  Foundation.
+ *
+ *  You should have received a copy of the GNU Lesser General Public License along with this
+ *  program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ *  or from the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ *  This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *  See the GNU Lesser General Public License for more details.
+ *
+ *  Copyright (c) 2006 - 2016 Pentaho Corporation..  All rights reserved.
+ */
+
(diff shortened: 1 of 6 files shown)


EgorZhuk:
Oh, need to make StateSequence Serializable

engine/core/source/org/pentaho/reporting/engine/classic/core/function/StateSequence.java
```diff
@@ -0,0 +1,90 @@
...
+
+package org.pentaho.reporting.engine.classic.core.function;
+
+import org.pentaho.reporting.engine.classic.core.states.ReportStateKey;
+import org.pentaho.reporting.engine.classic.core.util.BulkArrayList;
+import org.pentaho.reporting.engine.classic.core.util.Sequence;
+
+import java.io.Serializable;
+
+class StateSequence<T> implements Serializable {
```
tmorgner: Please make this class and its constructors public - it is a most helpful addition for anyone who wants to implement their own total-*-functions anywhere. OEMs love that stuff :)

EgorZhuk:
@tmorgner I've added a couple of junits for classfactory. Checked what is not covered in TotalGroupSumFunction - https://yadi.sk/i/DnOej471s8sJt Is it possible that global state key equals group key?

EgorZhuk:
Newline added

tmorgner:
> @tmorgner I've added a couple of junits for classfactory. Checked what is not covered in TotalGroupSumFunction - https://yadi.sk/i/DnOej471s8sJt Is it possible that global state key equals group key?

No. StateKey equality depends on the "stateCode" value. For global keys, it will be ReportEvent.REPORT_INITIALIZED, for group starts it will always be GROUP_STARTED. Under no (valid) circumstances can you run into that line.

tmorgner:
An easy way to test the TotalGroupSumFunction would probably be a mocked-up test. For the event, you will need a ReportState with a report and some data. You can use a dummy master-report with just some suitable groups to cover it. 

> RelationalGroup g = new RelationalGroup();
> g.setName("myGroup");
> g.addField("field-A");
>  MasterReport.addGroup(g);

The function only uses simple getters on the state, so you can get away with a very simple mock for it. DataRow has easy to use standalone implementations (StaticDataRow), so there is no pain here either.

tmorgner:
Good enough. Dont care about Wingmans test coverage randomness here.

tmorgner:
@EgorZhuk If your StateSequenceList is serializable with transient members, the list will have null for these members after deserialization. Is that really what you want here?

You probably need a readResolve(..) method here to restore the BulkArrayLists to some empty objects.

(serialization wont happen while the report is running, so restoring to an empty state here is fine.)

EgorZhuk:
@tmorgner nice catch, thank you, added it

PR #789, opened 2016-06-06
[BACKLOG-8501] - PIR: Sometimes message "The row limit has been reach…

dzmitry-hryhoryeu:
[BACKLOG-8501] - PIR: Sometimes message "The row limit has been reached" disappears when row limit is applied

Related pulls:
https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/373
https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/pull/435

@tmorgner
please review

### DIFF
--- a/engine/core/source/org/pentaho/reporting/engine/classic/core/layout/output/AbstractReportProcessor.java
+++ b/engine/core/source/org/pentaho/reporting/engine/classic/core/layout/output/AbstractReportProcessor.java
@@ -102,6 +102,11 @@ public abstract class AbstractReportProcessor implements ReportProcessor {
    */
   private boolean pagebreaksSupported;
 
+  /**
+   * A flag controlling whether query limit reached.
+   */
+  private boolean isQueryLimitReached;
+
   protected AbstractReportProcessor( final MasterReport report, final OutputProcessor outputProcessor )
     throws ReportProcessingException {
     if ( report == null ) {
@@ -422,7 +427,8 @@ public abstract class AbstractReportProcessor implements ReportProcessor {
 
       final ProcessState startState = new ProcessState();
       try {
-        startState.initializeForMasterReport( report, processingContext, (OutputFunction) lm.getInstance() );
+        boolean isQueryLimitReached = startState.initializeForMasterReport( report, processingContext, (OutputFunction) lm.getInstance() );
+        setQueryLimitReached( isQueryLimitReached );
       } finally {
         activeDataFactory = startState.getProcessHandle();
       }
@@ -1497,5 +1503,13 @@ public abstract class AbstractReportProcessor implements ReportProcessor {
    */
   private boolean isPagebreaksSupported() {
     return pagebreaksSupported;
+  }
+
+  public boolean isQueryLimitReached() {
+    return isQueryLimitReached;
+  }
+
+  public void setQueryLimitReached( boolean queryLimitReached ) {
+    isQueryLimitReached = queryLimitReached;
   }
 }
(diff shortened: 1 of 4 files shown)


tmorgner:
Two things: Copyright year, and the new code needs unit-tests.

dzmitry-hryhoryeu:
Thanks for reply! Copyright year changed and unit-test added.

tmorgner:
This code actually will not work as you'd expect. 

First, you are not able to tell the difference between exceeding the limit (you should warn), or just returning the same numbers of rows as the limit (you should not warn).

The reporting engine will try to push the query-limit downwards to the query itself and will allow the data-source to enforce the limit itself (ie via SQL's SELECT ... LIMIT x). If that fails, the reporting engine itself will then enforce the query-limit in the DefaultFlowController ( DefaultFlowController.java:250 ). 

So under no circumstances do you ever get a limit that is larger than the defined limit. 

Second: The query limit can be a computed value. The reporting engine will evaluate the attribute-expression that are set for these attributes and if they produce a valid result, it will use that result instead of the static value. 

I don't believe you can use PIR's approach here to create a more general implementation. 

To safely implement this feature, you will need to perform these checks in the DefaultFlowController's "performQueryData" method. To detect whether you exceeded the number of rows, perform your query with a limit that is 1 row larger than the requested limit. Then you can check:

     if (tablemodel.length >= (requestedLimit + 1) {
        set warning flag.
        use LengthLimitTableModel to reduce row-count to requested limit. 
     }
     else { 
       no limit was hit. Proceed as normal.
     }


There is no need to separate between hitting the user-limit or the system limit at this level. 

When you invoke the report, you can give the reporting engine a single number as limit. Just compute both the system-limit and the user-limit and provide the engine with the lower of the two. The reporting engine then just needs to tell you whether you did hit the limit at all. In your calling code you can then simply print "system" or "user" based on which limit value was lower. 

This moves the UI decision out of the engine-level code. It also means that you wont need to use the "UserQueryLimit" property at all. (I think those getter and setters can be eliminated when we clean up the various query-limit problems inside PIR.)

tmorgner:
On how to inform the caller that the query limit has been reached? 

Use a boolean return value on "ProcessState#initializeForMasterReport". You can then store that return value in your "AbstractReportProcessor" and make it available to the world via a getter.

You will get the value for it after the master-report query has been fired in ProcessState#initializeForMasterReport: Line 320++). You will have to add a boolean-property for that flag to DefaultFlowController to get that information out of that class, but that is a small price to pay.

dzmitry-hryhoryeu:
I made a change, according to your recommendations.

tmorgner:
In the test that runs reports, you can be a lot faster by simply using an empty master-report with a table-datafactory and an empty DefaultTableModel. Just give your model the required number of rows (via one of the available constructors). 

This way the report does not layouting (which is not needed in your case) and you get the perfect test scenario where you can control how many rows your query returns. 

And finally test the border conditions: TableModel with rows == query limit, rows < query limit and rows > query-limit. This will validate you whether you signal the correct result in each of the instances. This way your tests become truely meaningful and not just an hurdle to jump over for code coverage.

tmorgner:
I would also argue that if your query returns the same number row rows as in the query limit, you have not breached the limit and therefore the warning should not appear. That is why PIR's code always fired queries with "querylimit + 1".

dzmitry-hryhoryeu:
Сhanged test.

tmorgner:
Checkstyle and License headers, and then you are good to go.

dzmitry-hryhoryeu:
I Changed Checkstyle and License headers.

PR #791, opened 2016-06-21
[BACKLOG-8501] - PIR: Sometimes message "The row limit has been reach…

dzmitry-hryhoryeu:
[BACKLOG-8501] - PIR: Sometimes message "The row limit has been reached" disappears when row limit is applied

@tmorgner
please review


tmorgner:
Shouldn't there be tests to verify your change? You apparently work on a border case, so the junit tests should exist that validate that case. If there are existing unit tests and they did not fail until now, then those tests dont cover this particular case.

dzmitry-hryhoryeu:
Previous tests dont cover the situation when data source enforce limit itself.

tmorgner:
There are quite a few unused imports in the test file. Can you tell your IDE to clean it up please? 
After that we'll ignore wingman for a while and merge it. Poor chap has an identity crisis and it will be a while before he's back assisting our build validation.

dzmitry-hryhoryeu:
Unused imports was removed.

tmorgner:
Wingman, I love you, but you really need a therapy session.

PR #794, opened 2016-07-18
BACKLOG-7075 - Add custom interruption flag to Abstract report proces…

EgorZhuk:
…sor. Store processor in ThreadLocal.

@tmorgner need your opinion on this

### DIFF
--- a/engine/core/source/org/pentaho/reporting/engine/classic/core/layout/output/AbstractReportProcessor.java
+++ b/engine/core/source/org/pentaho/reporting/engine/classic/core/layout/output/AbstractReportProcessor.java
@@ -107,6 +107,12 @@ public abstract class AbstractReportProcessor implements ReportProcessor {
    */
   private boolean isQueryLimitReached;
 
+  /**
+   * A flag controlling report interruption
+   * Represents custom interruption mechanism not affected by switching thread interrupted flag
+   */
+  private boolean manuallyInterrupted;
+
   protected AbstractReportProcessor( final MasterReport report, final OutputProcessor outputProcessor )
     throws ReportProcessingException {
     if ( report == null ) {
@@ -153,6 +159,7 @@ public abstract class AbstractReportProcessor implements ReportProcessor {
 
     final boolean designtime = outputProcessor.getMetaData().isFeatureSupported( OutputProcessorFeature.DESIGNTIME );
     this.report = report.derive( designtime );
+    ReportProcessorThreadHolder.setProcessor( this );
   }
 
   protected ProcessStateHandle getProcessStateHandle() {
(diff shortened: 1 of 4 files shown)


EgorZhuk:
Connected with https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/395

tmorgner:
Can you hunt down the cases where we call "Thread.interrupt()" and make sure that we use the new cancellation mechanism instead? You most likely will find them in engine/core and the report-designer/report-designer-core sub-modules.

EgorZhuk:
@tmorgner fixed checkstyle, added junits. Unfortunately I can't find any inerruption code conneted with report processing in the project.

tmorgner:
Have a look at 
org.pentaho.reporting.engine.classic.core.modules.gui.base.PreviewPane#killThePaginationWorker

There is a call to "Worker.finish()" which then triggers an interrupt. The same code path is used in report-designer to cancel the pagination when you switch from edit-mode to preview and back.

Replace the "paginationWorker.finish()" with "printReportProcessor.cancel()" and there is one less interrupt in the world.

The normal exports seem to be non-interruptable (there is no cancel button on the dialog) so that should cover it all then.

tmorgner:
Egor, ReportProcessorThreadHolder does not exist. Did you forget to commit the file?

EgorZhuk:
Refactored test with InterruptedException.

EgorZhuk:
@tmorgner Could you please check if it is good enough now?

engine/core/source/org/pentaho/reporting/engine/classic/core/util/Worker.java
```diff
@@ -113,7 +115,10 @@ public void finish() {
       state = Worker.STATE_DEAD;
 
       try {
-        this.interrupt();
+        final AbstractReportProcessor processor = ReportProcessorThreadHolder.getProcessor();
```
tmorgner: I would rather see the cancellation code moved upwards to the PreviewPane. Technically either location would work, but conceptually, if you see a worker, you'd expect it to be not bound to a specific runnable type. Just use "printReportProcessor.cancel()" within the PreviewPane and your cancellation request will be granted.

Within the PreviewPane, you have direct access to the report-processor, so you don't even need to use the ReportProcessorThreadHolder for the cancellation. This makes the intent of the code much clearer. 

Another advantage (unnoticeable to the ordinary user): You now no longer actually kill the worker (state = STATE_DEAD), so the thread will be reused for the next task.

EgorZhuk:
@tmorgner let's see wingman's report

EgorZhuk:
@tmorgner seems tests are not broken

tmorgner:
Love it, merge it, and everyone lives happily ever after.

EgorZhuk:
@tmorgner  thank you, I will wait for the build on jenkins and will reopen PR to reporting plugin

PR #798, opened 2016-07-27
[PPP-3536] - Updated groovy to version 2.4.7 due to CVE-2015-3253

dkincade:
security vulnerability CVE-2015-3253 for groovy before 2.4.4

### DIFF
--- a/engine/core/ivy.xml
+++ b/engine/core/ivy.xml
@@ -56,7 +56,7 @@
 
       <dependency org="bsf" name="bsf" rev="2.4.0" transitive="false" conf="default_external->default"/>
       <dependency org="org.beanshell" name="bsh" rev="1.3.0" conf="default_external->default"/>
-      <dependency org="org.codehaus.groovy" name="groovy" rev="2.4.7" conf="default_external->default" transitive="false"/>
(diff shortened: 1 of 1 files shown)


tmorgner:
@dkincade Ups, you broke the build. Probably missing groovy-jsr223. You may be breaking customer scripting code if the new goovy-jar misses some previously included extra modules.

tmorgner:
The old groovy jar was a monolithic blob containing all extensions, I believe. I remember wanting to update that a long time ago, then seeing the "fun" involved (especially as this jar is used all over the place in the platform and XUL) and reconsidered my options and sanity. :D

Once finished, we can close PRD-4342 (Upgrade PRD to latest releases of Groovy).

dkincade:
@tmorgner Any objection to moving to groovy-all? that is the item being used in other places (like shims)

tmorgner:
No, I think it is safe.

tmorgner:
I no longer care about JDK 1.2 or bare-minimum below 10 megabyte downloads :)

dkincade:
After update to groovy-all ... looks good. I am going to include it.

PR #800, opened 2016-08-03
[BACKLOG-9531] - Added new option 'mime-message/text/html' to output-fo…

EgorZhuk:
…rmat attribute

@pentaho/orlandocalisbon @pamval please review

PR #801, opened 2016-08-04
[BACKLOG-9531] - Add email output to the publish dialog

EgorZhuk:
Allow user to select Email output from the publish dialog


EgorZhuk:
@tmorgner  please review

EgorZhuk:
@pamval @pentaho/orlandocalisbon please review

EgorZhuk:
@tmorgner could you please take a look?

tmorgner:
This is only half the fix. To make the email output appear in the "preferred-output-type" attribute, you will need to add it as extra entry to the org.pentaho.reporting.engine.classic.core.metadata.propertyeditors.ReportProcessTaskPropertyEditor as well. A "validValues.add("mime/text/html");" or so in the constructor will make it show up there as well.

EgorZhuk:
@tmorgner The second part is merged in #800

PR #802, opened 2016-08-08
[BACKLOG-6818] - Text rotation feature

EgorZhuk:


engine/core/source/org/pentaho/reporting/engine/classic/core/style/TextRotation.java
```diff
@@ -0,0 +1,106 @@
...
+import org.pentaho.reporting.engine.classic.core.util.ObjectStreamResolveException;
+
+import java.io.ObjectStreamException;
+import java.io.Serializable;
+
+public class TextRotation implements Serializable {
+  public static final TextRotation D_90 = new TextRotation( "90" );
+  public static final TextRotation D_270 = new TextRotation( "-90" );
+  private static final String CSS =
+    "transform: rotate(#deg); -ms-transform: rotate(#deg); -webkit-transform: rotate(#deg);";
```
tmorgner: The use of # as magic constant might store trouble for future maintainer. Replace # with {0} and use "MessageFormat.format(CSS, String.valueOf(getNumericValue()));" and you never have to explain the meaning of # in this context to any other developer.

EgorZhuk:
More checkstyle.

EgorZhuk:
@tmorgner is it OK, or more tests are needed?

tmorgner:
The coverage changes are all over the place, so I attribute it to the normal randomness of whatever happens inside the other tests. Coverage is good enough for me. Lets merge ...

PR #803, opened 2016-08-09
[BACKLOG-9099] - Add meta-inject plugin dependency

EgorZhuk:
@pamval please review


EgorZhuk:
Wingamn has taken a vacation for today...

pamval:
@mdamour1976 @wseyler @e-cuellar  @bmorrise @ddiroma @tgf @AlexanderBuloichik @AndreyBurikhin can you review and figure out if this is the correct approach?

Also, can you get one of those handy team tags so I can just tag @pentaho/r2d2 instead of tagging seven people?

bmorrise:
@pamval This should be a good short term fix. I'll look into this further today and see if there is something better long term.

PR #806, opened 2016-08-17
[BACKLOG-10072] - Some fields become null after rotation

EgorZhuk:
@tmorgner please take a look

### DIFF
--- a/engine/core/source/org/pentaho/reporting/engine/classic/core/filter/types/RotatableText.java
+++ b/engine/core/source/org/pentaho/reporting/engine/classic/core/filter/types/RotatableText.java
@@ -26,10 +26,12 @@ public interface RotatableText {
 
   default Object rotate( final ReportElement element, final Object value ) {
 
-    final TextRotation rotation =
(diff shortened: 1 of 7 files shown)


engine/core/source/org/pentaho/reporting/engine/classic/core/filter/types/RotatableText.java
```diff
@@ -35,4 +35,15 @@ default Object rotate( final ReportElement element, final Object value ) {
     return value;
   }
 
+
+  default Object rotate( final ReportElement element, final Object value, final Object defaultValue ) {
```
tmorgner: value is always non-null, except when querying fieldnames ("queryFieldName") or "element.getElementType().getMetaData().getName()". 

Wouldnt it be clearer to ensure that null does not happen in these 2 places there, for instance by checking the value beforehand? this way you dont need two methods that are almost completely the same. An "Assert.notNull(..)" at the start of the other rotate method will ensure that no one ever passes in null to start with.

tmorgner:
Looks good. There should be an easy win to gain Wingmans favours by restoring the levels of coverage for RotatableText and AbstractElementType.

PR #807, opened 2016-08-17
[BACKLOG-10064] - Configuration for rotated text as images

EgorZhuk:
@tmorgner please review


EgorZhuk:
@tmorgner I will check if it is available in File->Report Configuration dialog

EgorZhuk:
@tmorgner yes, it is

PR #808, opened 2016-08-17
[BISERVER-12951] Aggregators and group by are not applied to fields on PRD report that uses metadata datasource is running on server

mr9d:
- Role names in metadata.xmi were changed from Admin to Administrator


mr9d:
@tmorgner @mkambol @mdamour1976 please review.

tmorgner:
Non code change seems fine, even though Wingman chokes on some unrelated matter. It probably just wants a hug.

PR #809, opened 2016-08-18
[BACKLOG-10063] - Configuration to force writing all pages in paginat…

EgorZhuk:
…ed HTML

@tmorgner pleasse merge


tmorgner:
This setting is never used anywhere outside the BI-Server plugin (and there it is only used in a very specific setting). I'd say: Move it there, and keep it private there under the prefix "org.pentaho.reporting.platform.plugin".

PR #814, opened 2016-08-25
[BACKLOG-10061] - Ignore rotation for CSV and TXT outputs

EgorZhuk:
@tmorgner please review


EgorZhuk:
fixed test and styles

tmorgner:
Can you bring in a targeted RotatableText test to cover the new code. This way we keep the coverage on that class up.

 Other than that: Code looks fine to me.

EgorZhuk:
@tmorgner added test, also made RotatableText more safe

PR #817, opened 2016-09-08
[BACKLOG-10064] - improve html rotated text output

prakapenka:
When using css styles html rotate text from center by default. Change this to be more predicable: iff 90 rotate by left bottom, if -90 use left top corner. Prefer simple css over complex style- we have bands and other elements to align text in center or whatever required. Otherwise  - we have compute parent container size which is specified in pt points - means we can't determine pt->px. I would say here using java script could be a solution, but I assume we don't write inline javascript.


tmorgner:
You can use "transform-origin" for repositioning your rotation point. This is a relative measure, defaults to 50%. So set it to either 0% or 100% to rotate on the left or right edge of the text.

http://www.w3schools.com/cssref/css3_pr_transform-origin.asp

prakapenka:
wrong push close

PR #818, opened 2016-09-14
[PRD-5637]-Running PRPT on PRD throws "ORA-01000: maximum open cursors

denisprotopopov:
exceeded" error in Oracle DB
 - added use CachedRowSet. It let doesn't keep open ResultSet for each
    statement
 - fix copyright comments


denisprotopopov:
@mchen-len-son @pamval please review the changes

tmorgner:
Instead of relying on a vendor specific class (com.sun package), which may not be available in alternative implementations of the JDK, use the standard API to create a RowSet:

http://docs.oracle.com/javase/8/docs/api/javax/sql/rowset/RowSetProvider.html

denisprotopopov:
@tmorgner i have fixed it (..use the standard API to create a RowSet)

PR #820, opened 2016-09-14
[BACKLOG-10473] - fix styles for 1+ excel sheets.

prakapenka:
Refactor HSSFCellStyleProducer using builder to be more testable.
In advance it would be better to have HSSFCellStyleKey and CellStyle crated in one place.
BACKLOG-10473 initially was introduced by addition of rotation property we came keys not matching according styles. Now should be corrected.


engine/core/source/org/pentaho/reporting/engine/classic/core/modules/output/table/xls/helper/HSSFCellStyleProducer.java
```diff
@@ -590,9 +590,28 @@ public CellStyle createCellStyle( final InstanceID id, final StyleSheet element,
...
-    if ( element != null ) {
+    ExcelCellStyleBuilder builder = new ExcelCellStyleBuilder( this.workbook );
+
+    builder.withRotation( element );
+    builder.withElementStyle( element, styleKey );
+    builder.withBackgroundStyle( bg, styleKey );
+
+    final CellStyle hssfCellStyle = builder.build();
+
+    /*final CellStyle hssfCellStyle = workbook.createCellStyle();
```
tmorgner: Dont comment out code, delete it. If we need to look at the old code, we can look at Git's history.

tmorgner:
And it needs a code-style fix to make Wingman happy.

PR #824, opened 2016-09-23
[BACKLOG-11117] - Report Viewer: PentahoDatasourceConnectionProvider.…

dzmitry-hryhoryeu:
[BACKLOG-11117] - Report Viewer: PentahoDatasourceConnectionProvider.ERROR_0002 when changing parameters.


dzmitry-hryhoryeu:
@tmorgner please review

tmorgner:
Blindly closing the connection will backfire in a very bad way. The "getReferencedFields" method is also called while the report is processing (when sorting, for instance), so you cannot simply close the connection. Closing will close the statements and resultsets as well, causing you to loose the transaction context, and kills all cursors. Serious bad stuff!

The smarter fix is to check whether the connection was already open (is the "connection" variable null) and if so, dont close it. Only close it if this call opened the connection.

And then the fix is incomplete: The MDXDataFactory also opens a connection, and needs the same treatment. 

And last but not least: Where is the unit test that validates the newly fixed behaviour change for this code?

dzmitry-hryhoryeu:
@tmorgner please review

PR #826, opened 2016-09-27
[PRD-5746] Border radius values are saved under an incorrect key

Advent51:
-fixed


Advent51:
@mdamour1976 please review

mbatchelor:
Please create a test case that covers the code in question.

Advent51:
@mbatchelor Marc, I think there was a wingman's mistake. I can't find any test classes which were affected by my changes. May be new wingman's check remove this problem.

tmorgner:
@Advent51 There might never have been a unit test case for that. 

You can write a new test easily by defining a report with an element that has a border-radius set. Then save the report and reload it, and check that your loaded report contains the correct property values in the expected places. 

Use org.pentaho.reporting.engine.classic.core.testsupport.ReportWritingUtil#saveAndLoad to quickly save and immediately load a MasterReport object.

Advent51:
@tmorgner @mbatchelor test was added

Advent51:
@wingman-pentaho I guess there are some problems with wingman

PR #829, opened 2016-09-28
[backlog-10392] - parameter localizations engine/core

prakapenka:
separated engine/core functionality


prakapenka:
@tmorgner pls review

prakapenka:
wingman show no mercy, as usual.

prakapenka:
Let check what can you say now, WINGMAN!

prakapenka:
one more turn

tmorgner:
Your code needs a lot of manual work to get it right. OEMs wont like this very much. 

You can greatly reduce the verbosity of your code, your testing efforts and your later changes to the reporting plugin by adding the translation to the ParameterDefinitionEntry interface:

    public String getTranslatedParameterAttribute( final String domain,
                                                   final String name,
                                                   final ParameterContext parameterContext ) {
      final String translateId = getParameterAttribute( ParameterAttributeNames.Core.NAMESPACE,
              ParameterAttributeNames.Core.TRANSLATE_RESOURCE_ID,
              parameterContext );
      final String parameterAttribute = getParameterAttribute( domain, name, parameterContext );
      if (translateId == null)
      {
        return parameterAttribute;
      }
  
      try {
        final ResourceBundle resourceBundle = parameterContext.getResourceBundleFactory().getResourceBundle( translateId );
        if ( resourceBundle.containsKey( parameterAttribute ) ) {
          return resourceBundle.getString( translateId );
        }
        return parameterAttribute;
      }
      catch ( MissingResourceException | ClassCastException  e){
        return parameterAttribute;
      }
    }

There is no need for extra methods for each translatable entry. Just replace all usages of translatable attributes with this new method and your code will work as it should. 

For ListParameters, you also have only one single method that you need to change:
org.pentaho.reporting.engine.classic.core.parameters.DefaultListParameter#getValues

This returns a ParameterValues object. The easiest way of getting everything nicely testable would be to implement a "TranslatedParameterValues" object that simply delegates to the original ParameterValues implementations that the method normally returns. Give this translator the resource-bundle that you looked up from the parameter-context, and then translate the result of the "getTextValue(..)" method. 

This new class will be trivial to test, and now your list-parameters will automatically produce translated content if there is a resource-bundle and a translation for the value. 

The TranslatedErrorMessage is currently not used anywhere (other than in tests). That the error message will be a user-editable message that will be used during the parameter-validation. The majority of error messages that are generated in the DefaultReportParameterValidator are technical ones (object is wrong instance, not an array etc.). 

The only message that makes sense replacing would be the "DefaultReportParameterValidator.ParameterIsInvalidValue" and the "DefaultReportParameterValidator.GlobalError" message. These two indicate that the user made a input error other than leaving the field blank (which will be covered by a "this field is mandatory" message).

As before, I would simply add the necessary logic to the "DefaultReportParameterValidator" class so that it returns already translated errors. This reduces the need for duplicate changes everywhere else.

prakapenka:
1) We are not going to translate any arbitrary parameter attribute. Even more - we going to translate only label, display value, data format and error message - nothing else. Because something else can occasionally find a key in resource bundle for translation it is not supposed to be translated. What if we have numerous display values translations? Resource id is not only the predicate to translate something - we have to have explicit true for translating label from resource id etc.

2) Error message is not used in report designer - but this is a problem of report designer. What we need is ability to use custom error message per parameter. We have reporting plugin - and error messages are designed for it. Yes we have error messages from validator. But it is a server side validator - we going to validate parameters on a client as much as we can. What if user want custom error message for parameter? One of ways to use post-proc formula (and parameter becomes server-side-validated automatically), but will that value be translated? Is it so easy approach to use post proc formula to show error? What if I want to use post proc formula for something else? Even if somebody want to change validator messages, where should he look for properties files, inside a jar somewhere under web-inf/classes? The error message value we have defined for client side parameter validation, and if we want to - that will be translated error message. If it is server-side validated parameter - it will be validated on a server hopefully we have on a server (again - inside a jar in current implementation) translated value for user locale.

So maybe I would give this code a second chance.

tmorgner:
For (1) - I know that we only translate a few select properties. That does not prevent you from removing duplicate code or to bring code to where it belongs.

Your approach of introducing a new set of properties that mirror the non-translated values introduces redundancy that will come back to bite you or other developers later. You now have to manage two sets of values and either clear them when not needed or allow users to edit both (or you end up with garbage) stored in the report-files. 

You will be using *either* "label" or "translated-label" etc., but never both. You already know whether your label contains a translated entry or a non-translated entry by simply  looking at the parameter's TRANSLATE_RESOURCE_ID attribute. If it is filled, you are translating, if not, you are not translating. So having extra entries is redundant - you know the meaning of these values already. 

Btw: The values for the new constants like TRANSLATE_RESOURCE_ID should follow the style of the existing properties and use dashes instead of underscores. Keep the API consistent, and these values form part of the API, as they are exposed to the world in the PRPT file format and the parameter-xml on the server.

As for the "Parameter-helper":
All the inputs for your translation methods are held on the parameter-definition object. Therefore, object-orientated design guidelines tell us that there is a great likelyhood that your code suffers from "Feature envy" (Google it!) - it should be moved to where all your data is held. 

As a hint: If you find yourself making many complex changes in too many places, then treat it as a warning sign. 

Think forward on how you would use your changes to alter the parameter-editor in PRD, or the server's parameter handling. If your code here does not make these changes as easy as possible, then fix it before it becomes a problem later.

Another smoke test for the quality of an API is: Can you easily explain and document how to use your API to a fellow developer? Will there be special cases and traps he needs to be aware of? If you answer yes to any of the last two questions you have a red flag that requires attention. 

(2) Parameter validation error messages are produced by the DefaultReportParameterValidator. That is true for both the report-designer and the reporting plugin. If you make behaviour changes on the server, I would expect that the server and PRD behave in a similar way. If error messages are translated on the server, PRD should show the same translations when you test the report locally. Therefore the translation should be processed where the report parameters are validated - in the DefaultReportParameterValidator. 

The error messages the DefaultReportParameterValidator produces are normal strings and they are already translated. So if you want other text to show up when parameter are validated, you *have* to change it there - as a API user has no safe way to get to the original error code from the translation. 

In now also means that users of your ReportParameterValidator automatically get translated values when they are available - they dont have to change any of their code to make it work.

As for formula errors: At the moment, we have the "ERROR" formula function to produce custom errors. The error text is (in the current implementation) then added to a standard error message. That behaviour has to change. If the error is of type "CustomError", the text has been provided by the user and should be returned as it is, without any standard prefix before it. 

We then also need a new Formula-Function "RESOURCELOOKUP(bundleId, resourceId)" that can lookup a value from the resource-bundles of the report. That allows users to write their own error logic like this:

   IF ([value] = "invalid"; ERROR(RESOURCELOOKUP("myBundle", "myTranslation")); [value])

or if they want to format the message

   IF ([value] = "invalid"; ERROR(MESSAGEFORMAT(RESOURCELOOKUP("myBundle", "myTranslation"); [value]); [value])

In any case: For formulas it is the users responsibility to return a translated error value from their formula, but we have to make it possible to do so.    

The RESOURCELOOKUP function must be created in the engine/core project, as you need access to the "ReportFormulaContext" to access the resource-bundles of the report.

Hint: Formula error messages are handled in the class "FormulaParameterEvaluator". 


This "bolt on" coding style you use in your original commit produces very fragile code as you now have to change every occasion where parameter are presented to or edited by the user. In the case of error messages, there is no way you can add them without changing the parameter-validator itself. 

And again, we do have OEMs, they do embed our technology, and they can reasonably expect an API that is internally consistent and shows consistent behaviour.

prakapenka:
That is sound reasonable - especially for 'future envy' approach. May be that is my private feeling on 'if you want to edit existed public interface - try to edit something else'. Introducing `getTranslatedParameterAttribute`
will encapsulate retrieve of translated value to a parameter definition, so we will have 
`getTranslatedParameterAttribute + getParameterAttribute` 
both.

Other is not. TRANSLATE_RESOURCE_ID - is unique per parameter. That means we can use different properties file for every different parameter. I think that would be usefull if we are going to translate display values - (a dozen of display values) - and want to isolate it in one file. Also - for this parameter we are going to translate a lablel. Or we want to translate label and do not translate display values. What if we will end up when one of display names is equal to label name -> will be translated as a label. From my point of view this is WTF situation and I would aware of it. That means everywhere we going to display label (as far as trying to format date as far as translate display values) we have to check we are really going to translate it. That means we came to edit every place where do we extract this values.

I'll try to clarify it once again - if 'TRANSLATE_RESOURCE_ID' is enough precondition for translation - take a look into - BACKLOG-10392 screenshots - we don't need check-boxes to translate lablel/etc. - if resource id is specified -> we try to translate all values we will find as a keys in resource lookup.

 As for 'RESOURCELOOKUP' function - this definitely will alter error message field. And this will be the case when - if you want custom error message - you want server side validation. You can write custom error handler for every parameter - and get messages from server validation.

Am I understand this correct?

tmorgner:
Adding a new method to an interface is totally fine when you introduce a new feature. We are on a major release right now, so changing interfaces is the sane thing to do. If we were in a maintenance release (7.1 for instance), you'd have to think about binary compatibility. But we are on Java8, so a default implementation on the interface would fit that bill.

In any case: I prefer maintainability (and thus implicitly easy of use) over anything else - the code we write today is going to be around long after we are gone, so we better make it as safe and sane as we possibly can. Future maintenance engineers will thank you for it.


During the grooming, when we were talking about translating display-values we were talking about having a separate resource-bundle id for the display-values. These translations will most likely be shared across several reports, so it makes sense to allow them to be reusable. I would solve it by simply having a second attribute "translate-display-values-resource-id". If a user wants to use the same resource-bundle in both places, they can do so easily by entering the same id twice, and if they have a centrally managed one, they can do that too. Win win!

PR #832, opened 2016-10-03
[PRD-5757]-Results sets with booleans are not cached

denisprotopopov:
- added check for Booleans
 - added unit test


tmorgner:
Sweet find! Merged.

PR #833, opened 2016-10-03
[backlog-10392] - parameter localizations engine/core

prakapenka:
Introduce new formula for resource lookup. We may consider to review ResourceLookupFunction.properties  for better display values.

### DIFF
--- a/engine/core/test-src/org/pentaho/reporting/engine/classic/core/modules/gui/base/parameters/TextFieldParameterComponentTest.java
+++ b/engine/core/test-src/org/pentaho/reporting/engine/classic/core/modules/gui/base/parameters/TextFieldParameterComponentTest.java
@@ -12,33 +12,27 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2000 - 2015 Pentaho Corporation, Simba Management Limited and Contributors...  All rights reserved.
+ * Copyright (c) 2000 - 2016 Pentaho Corporation, Simba Management Limited and Contributors...  All rights reserved.
  */
 
 package org.pentaho.reporting.engine.classic.core.modules.gui.base.parameters;
-
-import static org.hamcrest.CoreMatchers.equalTo;
-import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertThat;
-import static org.mockito.Matchers.any;
-import static org.mockito.Mockito.doReturn;
-import static org.mockito.Mockito.mock;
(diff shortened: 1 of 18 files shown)


prakapenka:
Et dices ad Wingman

prakapenka:
hey wingman check it again

engine/core/test-src/org/pentaho/reporting/engine/classic/core/modules/gui/base/parameters/TextFieldParameterComponentTest.java
```diff
@@ -38,7 +38,6 @@
 import org.junit.Before;
 import org.junit.Test;
 import org.pentaho.reporting.engine.classic.core.ResourceBundleFactory;
-import org.pentaho.reporting.engine.classic.core.modules.gui.base.ParameterReportControllerPane;
 import org.pentaho.reporting.engine.classic.core.parameters.ParameterAttributeNames;
```
prakapenka: - import org.pentaho.reporting.engine.classic.core.modules.gui.base.ParameterReportControllerPane; is deleted while wingman still claim for

"/home/buildguy/project/engine/core/test-src/org/pentaho/reporting/engine/classic/core/modules/gui/base/parameters/TextFieldParameterComponentTest.javaorg.pentaho.reporting.engine.classic.core.modules.gui.base.ParameterReportControllerPane. at Line: 41 Column: 8"

wingman shows no mercy as usually.

prakapenka:
wingman, wingman, preaching wing...

prakapenka:
@tmorgner please review

tmorgner:
Wingman says there is work to do.

tmorgner:
How does this related to the other pull request #829 ? If that other one is obsolete now, remove it. 
If not, then you have overlapping changes and should bring both together or you get merge problems later on.

I also noticed that this PR here has no support for translating display-values of list-parameters. If that is intentional and you plan to add that later, then that is fine by me - just tell whether there will be another pull request or whether this PR will be extended.

So far, the code looks good. Nothing to complain - Wingman does a good job with that anyway :)

prakapenka:
No need to translate display names via DefaultListValues since it has ComputedListValues and having for display values post-processing formula - we will get the same result just with no changes at all. Since we ok to translate error messages via foumlas expression - seems we ok to translate display values with post proc foumula. If we would have formulas for data format and for display label - we would write only resourcelookup formula and nothing else. Now we are going to define resource for label and data format in separate input. There also will be second PR to update designer UI immediately after this PR will be merged.

prakapenka:
Wingman, the lord master of checkstyle purgatory, show me some mercy.

tmorgner:
Waiting for Wingman, then merging ..

PR #835, opened 2016-10-03
[PRD-5762] - PDI as a datasource in PRD. Fields not being read from t…

stanislau-strelchanka:
[PRD-5762] - PDI as a datasource in PRD. Fields not being read from transformation.

### DIFF
--- a/designer/report-designer-assembly/build.properties
+++ b/designer/report-designer-assembly/build.properties
@@ -18,6 +18,7 @@ dependency.jsonpath.revision=1.0
 
 dependency.pentaho-big-data-plugin.revision=7.0-SNAPSHOT
 dependency.pentaho-cassandra-plugin.revision=7.0-SNAPSHOT
+dependency.kettle-xml-plugin.revision=7.0-SNAPSHOT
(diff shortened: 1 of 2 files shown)


stanislau-strelchanka:
@wingman-pentaho can you please assist with errors

tmorgner:
@wingman-pentaho Isn't dev-lib supposed to be an optional directory? And the error points to line 2084 of subfloor, which is in master is a "taskdef" for the JUnit ant-task. It doesn't even mention "devlib" anywhere near there. Makes me wonder what subfloor.xml wingman is actually using.

stanislau-strelchanka:
@mkambol, @mdamour1976, @tmorgner, @mchen-len-son, @pamval  please review

stanislau-strelchanka:
@mkambol, @mdamour1976, @tmorgner, @mchen-len-son, @pamval, @wingman-pentaho 
'compile' target is used from 'build.assembly.sequential.xml' and it didn't depend on 'init' which
is responsible for creating devlib and other folders

please review

stanislau-strelchanka:
moved wingman fix to #840

PR #836, opened 2016-10-05
[BACKLOG-10064] - Text in rotated element exceeds placeholder borders.

dzmitry-hryhoryeu:
[BACKLOG-10064] - Text in rotated element exceeds placeholder borders.

@tmorgner please review.


tmorgner:
Just some unit tests would make this perfect. You can directly output a test report to HTML and validate that the element has the correct style there. 

The engine outputs XHTML, so you can directly validate the output as XML. If you use inline styles, then the engine delivers the CSS directly on to the element, which makes your validation extra easy. And if you give your report-element an unique ID, that ID will be set on the generated XML-element as well, which makes locating your generated content within the document easy too.

So there is no reason why this code does not have any unit-tests at all.

tmorgner:
@dzmitry-hryhoryeu Ping. Pull request needs attention.

dzmitry-hryhoryeu:
@tmorgner I am working on the unit tests.

dzmitry-hryhoryeu:
@tmorgner please review.

PR #837, opened 2016-10-05
[PRD-5766] Typo on code reads attribute as timout instead of timeout

Advent51:
-typos fixed


tmorgner:
Checkstyle complains, and why is there no unit-test?

PR #838, opened 2016-10-07
[PRD-5785]-Morphing Resource Labels clearing out SubReport

denisprotopopov:
- check the value for null, and remove the entry from the attribute-map


tmorgner:
Totally merged.

PR #842, opened 2016-10-13
[BACKLOG-11444] Pooled big-data connections fail

mkambol:
PooledDatasourceHelper.convert() in reporting had the same
issue as the comparable method in platform.  Since this one
is unused I'm removing it.


mkambol:
@tmorgner - removed .convert as discussed.  Coverage dropped some just because the unused code had tests.

tmorgner:
I can live with the dropped coverage. Deleting code is the ultimate debugging, just like burning down the house to kill the mosquito :)

PR #845, opened 2016-10-18
[PPP-3584] PRD - Security Vulnerability when using XML file as datasource

AliaksandrShuhayeu:
[PPP-3584] PRD - Security Vulnerability when using XML file as datasource

- disable default support DTD in pentaho report designer
- add property for enable/disable support DTD

### DIFF
--- a/engine/extensions-xpath/source/org/pentaho/reporting/engine/classic/extensions/datasources/xpath/LegacyXPathTableModel.java
+++ b/engine/extensions-xpath/source/org/pentaho/reporting/engine/classic/extensions/datasources/xpath/LegacyXPathTableModel.java
@@ -12,16 +12,18 @@
 * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Lesser General Public License for more details.
 *
-* Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+* Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
 */
 
 package org.pentaho.reporting.engine.classic.extensions.datasources.xpath;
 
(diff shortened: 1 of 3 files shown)


engine/extensions-xpath/source/org/pentaho/reporting/engine/classic/extensions/datasources/xpath/LegacyXPathTableModel.java
```diff
@@ -96,12 +104,20 @@ public LegacyXPathTableModel( final ResourceData xmlResource,
       columnTypes = new ArrayList<Class>();
       columnNames = new ArrayList<String>();
 
+      DocumentBuilderFactory df = DocumentBuilderFactory.newInstance();
+      final Configuration configuration = ClassicEngineBoot.getInstance().getGlobalConfig();
+      if ( !"true".equals( configuration
+              .getConfigProperty( "org.pentaho.reporting.designer.support.dtds.enable" ) ) ) {
```
tmorgner: Please change the property name to match the package of the XPAth datasource. It is not just report-designer but also BI-Server that are affected.

Use "org.pentaho.reporting.engine.classic.extensions.datasources.xpath.EnableDTDs" as property name, for instance.

tmorgner:
Just reading the OWASP guide: I would also add "dbf.setXIncludeAware(false);" - or I craft you an XInclude that kills your server.

AliaksandrShuhayeu:
@tmorgner I added dbf.setXIncludeAware(false); but it false by default and renamed property.

pamval:
@tmorgner I'm reverting this change since we want to do a risk/scope assessment before including this in 7.0. This will be merged to master after we branch 7.0 tomorrow

tmorgner:
Aye, no problem. There are other places that may need fixing as well.

PR #846, opened 2016-10-18
[BACKLOG-11620] Update product App icons and splash screens

pedrofvteixeira:
- with the change in the splash screen image from having a white background to a dark blue one, the color of the text needs to be set to white ( in opposed to default black )


pedrofvteixeira:
@mdamour1976 , @tmorgner , @pentaho/rogueone please review.

PR #848, opened 2016-10-19
[PPP-3584] PRD - Security Vulnerability when using XML file as datasource

AliaksandrShuhayeu:
[PPP-3584] PRD - Security Vulnerability when using XML file as datasource

- disable default support DTD in pentaho report designer
- add property for enable/disable support DTD

@pamval


pamval:
please wait for 7.0 to be branched before merging this

AliaksandrShuhayeu:
@pamval @mbatchelor 
I refactored my changes and added tests. 
For full correct testing LegacyXPathTableModel we need deep refactor this calss. 
Now it is not OOP it has actually one method (constructor) with full logic in it.

PR #850, opened 2016-10-20
[CLEANUP] - Update project definition to allow Java 1.8 features.

tmorgner:
Removed inspection profiles. There is no need for that file to be versioned and each IntelliJ release makes its own (bogus) changes here. Troublemaker!


tmorgner:
No code changes, so Wingman, I ignore thee.

PR #851, opened 2016-10-20
[BACKLOG-11732] - Expose locale information as fields to make sure we…

tmorgner:
… can use it as query parameter.


tmorgner:
@dkincade @kcruzada The missing piece for parameter localisation. (master version)

dkincade:
@tmorgner Please correct the license header and checkstyle violations in a separate [CLEANUP] commit.

PR #852, opened 2016-10-20
[BACKLOG-11732] - Expose locale information as fields to make sure we…

tmorgner:
… can use it as query parameter.

(cherry picked from commit b97e611)


tmorgner:
@dkincade @kcruzada The missing piece for parameter localisation.

Same as #851

PR #853, opened 2016-10-20
[CLEANUP] - Fixed copyright year and template; codestyle fixes

tmorgner:


tmorgner:
One should not clean out stale branches that cover an open pull request ..

tmorgner:
@wingman-pentaho @dkincade  The Wingman configuration for building this project is wrong. "engine/sdk" is an assembly project and has no valid "jacoco" target. Shall I fix the build, or do you want to fix Wingman's config.

If you run the build as "ant continuous-local" you get the projects all resolved and built in the right order, and Jacoco will run on those modules that can handle it.  Subfloor defines a standard "build" target that does the full compile and test-with-coverage cycle for you. This target is maintained by subfloor/the build-team and we should use it. 

Also: I stated that before: Your approach will not work if there are API changes across the modules, as Ant calls each command-line target independently. If you want to ensure the correct order, you have to call a single local target on a multi-build ant-file.

lucboudreau:
Wingman can't invoke different targets for different build files within the same repository. We need all build files within the repo (wether root level or subfolder) to use the same target.

If there was one common to all, what would it be, @tmorgner?

cc. @wingman-pentaho

tmorgner:
"build": It does what your command-line does manually. But historically, we have overrides in place on "build" for projects that had no build activities or that only needed a "dist" etc.

For the definition of "build" in Subfloor:
https://github.com/pentaho/subfloor/blob/master/master-copies/subfloor.xml#L508

This at least will fix this problem here. However, Wingman will still choke-up if ever there is a API change that propagates between modules. This is due to Ant's handling of multiple-targets on a command line. When you invoke "ant TargetA TargetB" on a multi-build target, you get the following sequence:

       parent: TargetA
         child0: TargetA
         childN: TargetA
       parent: TargetB
         child0: TargetB
         childN: TargetB

which if TargetA is resolve, and TargetB is publish-local, is all wrong. To fix this and to allow tools like Wingman to work nicely, I would recomment to add a new target to Subfloor called "continuous-local", that combines "build" and "publish-local" in the same way "continuous" combines "build" and "publish".

https://github.com/pentaho/subfloor/blob/master/master-copies/subfloor.xml#L547

PS: In an ideal world, we would all be on Maven now. But even if we move Master to maven first thing tomorrow, the 6.1 branch would still be an Ant based build. Fixing subfloor will fix all currently maintained branches with no side-effects.

lucboudreau:
@tmorgner 

We can already define submodules and their build order in Wingman. I could also build only the modules that we want and skip those that won't build. 

What is the name/path/order of the modules you want built for this project?

Our general aim w/ Wingman is to build as little as possible, while still doing it in the right order, should submodules exist. With Maven, obviously, this is pretty much done through metadata of the project itself (when properly formed and configured) but for Ant, we unfortunately need to define those manually and had to build our own logic.

tmorgner:
Here are all modules that should be built in their correct order

        libraries/flute,
        libraries/libbase,
        libraries/libpixie,
        libraries/libloader,
        libraries/libxml,
        libraries/librepository,
        libraries/libdocbundle,
        libraries/libfonts,
        libraries/libformula,
        libraries/libformat,
        libraries/libpensol,
        libraries/libswing,
        libraries/libsparkline,
        libraries/libformula-ui,
        libraries/libserializer,
        libraries/configuration-editor,
        libraries/libcss,
        engine/core,
        engine/extensions,
        engine/extensions-cda,
        engine/extensions-docsupport,
        engine/extensions-drilldown,
        engine/extensions-kettle,
        engine/extensions-mondrian,
        engine/extensions-olap4j,
        engine/extensions-openerp,
        engine/extensions-pentaho-metadata,
        engine/legacy-charts,
        engine/extensions-xpath,
        engine/extensions-reportdesigner-parser,
        engine/extensions-sampledata,
        engine/extensions-scripting,
        engine/extensions-toc,
        engine/legacy-functions,
        engine/wizard-core,
        engine/demo,
        engine/testcases,
        engine/samples,
        designer/wizard-xul,
        designer/datasource-editor-cda, 
        designer/datasource-editor-external, 
        designer/datasource-editor-jdbc, 
        designer/datasource-editor-kettle, 
        designer/datasource-editor-mondrian, 
        designer/datasource-editor-olap4j, 
        designer/datasource-editor-pmd, 
        designer/datasource-editor-reflection, 
        designer/datasource-editor-scriptable,
        designer/datasource-editor-table, 
        designer/datasource-editor-xpath,
        designer/datasource-editor-openerp, 
        designer/report-designer,
        designer/report-designer-extension-legacy-charts,
        designer/report-designer-extension-pentaho,
        designer/report-designer-extension-toc,
        designer/report-designer-extension-connectioneditor,
        designer/report-designer-extension-wizard,
        designer/report-designer-assembly

The following ones are not built:

        engine/sdk
        engine/tools

and for report-designer-assembly I would say it is up to you. It is a glorious "zip" job with no tests nor validation logic involved. Wingman wont detect any coding errors here. Include it if you want validation of the build process in general, but to keep it lightweight, I would setup a job that specifically monitors just the assembly-subdirectory for changes in GIT in that case. Downloading OSGI- and big-data bundles tests your bandwidth, but does not test much in relation to "normal" code-change commits.

lucboudreau:
I've created a [case](http://jira.pentaho.com/browse/ENGOPS-2521) to track this. 

For the assembly, we usually exclude anything related to assemblies in other repositories. So we will omit this one too.

dkincade:
The code changes for this PR look very good. Accepting.

PR #855, opened 2016-10-20
[CLEANUP] - Fixed copyright year and template; codestyle fixes. Delet…

tmorgner:
…ed "quality" inspections, they are modified by IntelliJ all the time and are not critical for the project definition itself.  Once deleted, this file will be covered by the gitignore rule.


tmorgner:
@wingman-pentaho Seems the Wingman configuration is wrong. "engine/sdk" is an assembly project and should not need "jacoco". It contains neither sources nor tests, so it can be skipped. 

If you want to build something here, then "dist" is the only sensible target to be called, which will produce a zip file.

PR #857, opened 2016-10-24
[PRD-5264]-PRD gets frozen when using CSVARRAY within the parameter d…

denisprotopopov:
…efinition

 - added addinional condition in hasMoreToken for case when separator empty
   string
 - added unit test


tmorgner:
Code looks good. Lets wait for Wingman's final seal of approval.

PR #863, opened 2016-10-26
[BACKLOG-11906] - fix formula computation for empty label and data format

prakapenka:
The tricky part is that if label of data format fields is empty, but formula are used - no attribute is set for label or data format. So having approach to add '-formula' will not help since we don't have such attributes at all.
In parameter dialog, as far as in bundle writers like [here](https://github.com/pentaho/pentaho-reporting/blob/497ca68305ea5183d8af0972aae339512791aae4/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/parser/bundle/writer/DataDefinitionFileWriter.java#L218) we will not save empty attributes.

So in case of label of formula fields are empty - we will just write formula value there. On UI if we found that label value is equal to formula value - we just set empty text field. When going to save - replace empty with formula value again.

### DIFF
--- a/designer/report-designer/ivy.xml
+++ b/designer/report-designer/ivy.xml
@@ -60,13 +60,14 @@
 
     <!-- testing dependencies -->
     <!-- Testing Dependencies -->
-    <dependency org="junit" name="junit" rev="4.10" transitive="false" conf="test->default"/>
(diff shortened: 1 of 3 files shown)


prakapenka:
Ok, no mockito dependency when running with ant as far as 'java.awt.headless' should be fixed now.

prakapenka:
Oh, forget about ctrl+shift+o. Should be ok now.

PR #864, opened 2016-10-27

YuryBY:
@mbatchelor, @mkambol, could you please take a look?

### DIFF
--- a/engine/extensions-mondrian/source/org/pentaho/reporting/engine/classic/extensions/datasources/mondrian/MondrianUtil.java
+++ b/engine/extensions-mondrian/source/org/pentaho/reporting/engine/classic/extensions/datasources/mondrian/MondrianUtil.java
@@ -12,7 +12,7 @@
 * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Lesser General Public License for more details.
 *
-* Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+* Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
 */
 
 package org.pentaho.reporting.engine.classic.extensions.datasources.mondrian;
@@ -24,6 +24,7 @@ import mondrian.olap.Util;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.pentaho.reporting.engine.classic.core.ClassicEngineBoot;
+import org.pentaho.reporting.libraries.base.util.XMLParserFactoryProducer;
 import org.pentaho.reporting.libraries.resourceloader.ResourceCreationException;
 import org.pentaho.reporting.libraries.resourceloader.ResourceKey;
 import org.pentaho.reporting.libraries.resourceloader.ResourceManager;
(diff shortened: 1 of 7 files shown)


engine/extensions-mondrian/source/org/pentaho/reporting/engine/classic/extensions/datasources/mondrian/MondrianUtil.java
```diff
@@ -66,8 +67,11 @@ public static String parseSchemaName( final ResourceManager resourceManager,
     return null;
   }
 
-  private static String parseXmlDocument( final InputStream stream ) throws ResourceCreationException {
+  private static String parseXmlDocument( final InputStream stream ) throws ResourceCreationException,
+      ParserConfigurationException {
     final DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+    dbf.setFeature( XMLConstants.FEATURE_SECURE_PROCESSING, true );
```
mbatchelor: This is in the same .jar file as DocumentHelper above and has the same access to XMLParserFactoryProducer as DocumentHelper above. So please have this call call into XMLParserFactoryProducer the same as DocumentHelper (and everywhere else). Ultimately, there will be some other defect that we will need to fix, and fixing it in one place for PRD is better than fixing it in 6.

libraries/configuration-editor/source/org/pentaho/reporting/tools/configeditor/util/DOMUtilities.java
```diff
@@ -62,6 +63,8 @@ private DOMUtilities() {
   public static Document parseInputStream( final InputStream instream )
     throws ParserConfigurationException, SAXException, IOException {
     final DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+    dbf.setFeature( XMLConstants.FEATURE_SECURE_PROCESSING, true );
```
mbatchelor: configuration-editor uses libbase everywhere. I'd rather this whole method went away in favor of using ConfigurationMetaDataParser.parseInputStream and make the libbase one a public static like this one. Then we have one parse thingie in libbase that everyone else can call.

libraries/libbase/source/org/pentaho/reporting/libraries/base/config/metadata/ConfigurationMetaDataParser.java
```diff
@@ -45,6 +46,8 @@
   private static Document parseInputStream( final InputStream instream )
```
mbatchelor: Suggest this goes to public static so that everyone else has one place to call to set up a secure parser.

libraries/libxml/source/org/pentaho/reporting/libraries/xmlns/parser/DomTreeResourceFactory.java
```diff
@@ -69,12 +70,13 @@ public Object getResource() {
...
                           final ResourceData data,
                           final ResourceKey context ) throws ResourceCreationException, ResourceLoadingException {
-
-    final DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
-    dbf.setNamespaceAware( true );
-    dbf.setValidating( false );
-
+    final DocumentBuilderFactory dbf;
     try {
+      dbf = DocumentBuilderFactory.newInstance();
```
mbatchelor: Suggest this part gets moved to libbase (the documentbuilderfactory.newInstance() and security stuff) so that we're moving that code all into a single place in the lib*.* world.

YuryBY:
@mbatchelor, have found this one... already updated. Thanks.

PR #865, opened 2016-10-27

YuryBY:
@mbatchelor, @mkambol, could you please take a look?


YuryBY:
@mbatchelor, here is correct one for 7.0: https://github.com/pentaho/pentaho-reporting/pull/893. Thanks.

PR #867, opened 2016-10-27
[SP-2955][PRD-5757]-Results sets with booleans are not cached

denisprotopopov:
- added check for Booleans
- added unit test


mchen-len-son:
@denisprotopopov please fix the checkstyle violation

PR #868, opened 2016-10-27
[SP-2939][PRD-5785]-Morphing Resource Labels clearing out SubReport

denisprotopopov:
- check the value for null, and remove the entry from the attribute-map


mchen-len-son:
@denisprotopopov don't you need to also backport the #823 PR?

denisprotopopov:
@mchen-len-son no i don't. PR #823 was reverted

mchen-len-son:
ok, then.
@tmorgner could you please take a look?

PR #872, opened 2016-10-28
PRD-5780 Chart Background not showing in Export to PDF - Insufficient data for an image

AliaksandrDrebenchuk:
added the Image conversion to TYPE_3BYTE_BGR type


AliaksandrDrebenchuk:
@tmorgner
@mkambol
@mdamour1976
please review

tmorgner:
Your fix is an accidental fix. You are re-encoding all incoming images as PNG. The fact that you strip out the Alpha-Channel is actually not a requirement for the fix. You are killing people's transparent images for no particular reason.

The PDF apparently contains a corrupted data stream (as Acrobat reader helpfully reports). So your first question should be to investigate why there is an corrupted stream. To do that, you need to be able to detect such streams, without the help of Acrobat Reader, as we want unit-tests. 

Take the following test as an example on how to use IText to read a PDF and to inspect the stream.

      @Test
      public void checkPdfImagesForCorruption() throws Exception {
        URL u = getClass().getResource( "Prd-5780.prpt" );
        final MasterReport report = (MasterReport) new ResourceManager().createDirectly( u, MasterReport.class ).getResource();
        ByteArrayOutputStream bout = new ByteArrayOutputStream(  );
        PdfReportUtil.createPDF( report, bout );
    
        PdfReader reader = new PdfReader( bout.toByteArray() );
        for (int i = 1; i < reader.getXrefSize(); i+= 1) {
          final PdfObject pdfObject = reader.getPdfObject( i );
          if (pdfObject.isStream()) {
            PRStream stream = (PRStream) pdfObject;
            if ( PdfName.IMAGE.equals( stream.get( PdfName.SUBTYPE ) ) ) {
              byte[] raw = PdfReader.getStreamBytesRaw(stream, reader.getSafeFile());
              if (PdfName.FLATEDECODE.equals( stream.get( PdfName.FILTER) ) ){
                Assert.assertNotNull( "Stream is corrupted.", PdfReader.FlateDecode( raw, true ));
              }
            }
          }
        }

Without the fix, this test will fail and proves that there are corrupt streams. Now proceed to fix it:

The corrupted image declares that it has a "/FlatDecode" filter applied. However, compare the bytes of the PDF stream with the raw-bytes as seen by the PixelGrabber used in "com/lowagie/text/Image.java:629" (use https://github.com/hwinkler/itext2 as reference). They are identical, so that is no FlatDecoded content here. 

Long story short: 
Look at: org.pentaho.reporting.engine.classic.core.modules.output.pageable.pdf.internal.PdfGraphics2D#drawImage(java.awt.Image, java.awt.Image, java.awt.geom.AffineTransform, java.awt.Color, java.awt.image.ImageObserver) (Line 1342) 

In there, there is a call "image.setDeflated(true);" which should not be there. Remove the two calls and run the report and test. The unit-test now passes, and visual inspection of the resulting PDF also shows the correct image. 


Please update the pull request to this fix.

PR #875, opened 2016-10-31
[SP-2951] Backport of PRD-5746 - Border radius values are saved under an incorrect key (6.1 Suite)

Advent51:
Backport of #826


Advent51:
It seems like those failing tests are non-deterministic ones - they are passing locally

PR #880, opened 2016-11-02
[BACKLOG-12084] Revert the default theme back to crystal

pedrofvteixeira:
- reverting icons, splash screens and splash screen text colour


pedrofvteixeira:
@dkincade , @mdamour1976 please review.

I'll issue a PR for 7.0.0.0 in follow-up to this one.
Additional PR's for the other client tools and the installer will be issued once ready.

CC'ing @pentaho/rogueone

dkincade:
We are not reverting the Crystal theme in anything except PUC

PR #882, opened 2016-11-10
6.1

jwalor:


tmorgner:
We never merge bug-fix branches back into master. If a bug is fixed, the change is made in both master and the bug-fix branches. Closing as an invalid pull request.

PR #886, opened 2016-11-21
Mavenizing pentaho-reporting

EgorZhuk:
All pom files are included in a sepparate commit.
See README to configure Mac and Win/Linux builds.


tmorgner:
Egor, can you change the pull request so that it leaves the Ant scripts untouched (at least for now)?

This will ensure that the builds elsewhere still work. Then the build team can go in, review the Maven files, even take a few days to update and test their build processes, and then give us to go ahead to remove the ant scripts when everything runs smooth.

I'd hate it if we were the cause of large scale build chaos.

EgorZhuk:
@tmorgner ant build won't work,  the directory structure was changed to match maven's defaults.   Nevertheless I can modify ant build to make it support new directory structure.

EgorZhuk:
Accordingly to ENGOPS guideline there is no need to keep ant build files because separate jobs will be created for the maven build. If something goes wrong it is not a big deal to build it with ant. So, we are ready to go.

PR #889, opened 2016-11-28
PRD-5776 Unable to open saved prpt report having Japanese surrogate characters

AliaksandrDrebenchuk:
org.owasp.encoder was included instead custom XML encoding

### DIFF
--- a/libraries/libxml/source/org/pentaho/reporting/libraries/xmlns/writer/XmlWriterSupport.java
+++ b/libraries/libxml/source/org/pentaho/reporting/libraries/xmlns/writer/XmlWriterSupport.java
@@ -12,7 +12,7 @@
 * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Lesser General Public License for more details.
 *
-* Copyright (c) 2001 - 2013 Object Refinery Ltd, Pentaho Corporation and Contributors..  All rights reserved.
+* Copyright (c) 2001 - 2016 Object Refinery Ltd, Pentaho Corporation and Contributors..  All rights reserved.
 */
 
 package org.pentaho.reporting.libraries.xmlns.writer;
@@ -23,6 +23,7 @@ import org.pentaho.reporting.libraries.base.util.StringUtils;
 import org.pentaho.reporting.libraries.xmlns.common.AttributeList;
 
 import java.io.IOException;
+import java.io.StringWriter;
 import java.io.Writer;
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetEncoder;
@@ -173,9 +174,6 @@ public class XmlWriterSupport {
   private boolean writeFinalLinebreak;
   private boolean htmlCompatiblityMode;
   private String lineSeparator;
-  private StringBuffer normalizeBuffer;
-  private String encoding;
-  private boolean encodingvalid;
   private CharsetEncoder charsetEncoder;
 
   /**
@@ -218,7 +216,6 @@ public class XmlWriterSupport {
       throw new NullPointerException( "LineSeparator must not be null" );
     }
 
-    this.normalizeBuffer = new StringBuffer( 128 );
     this.safeTags = safeTags;
     this.openTags = new FastStack();
     this.indentString = indentString;
@@ -325,9 +322,6 @@ public class XmlWriterSupport {
   }
 
   public void setEncoding( final String encoding ) {
-    this.encoding = encoding;
-    this.encodingvalid = true;
-
     final Charset charset = Charset.forName( encoding );
     this.charsetEncoder = charset.newEncoder();
   }
@@ -655,8 +649,7 @@ public class XmlWriterSupport {
       }
     } else {
       final ElementLevel level = (ElementLevel) openTags.peek();
-      if ( getTagDescription().hasCData
-        ( level.getNamespace(), level.getTagName() ) == false ) {
+      if ( getTagDescription().hasCData( level.getNamespace(), level.getTagName() ) == false ) {
         writeNewLine( w );
       }
     }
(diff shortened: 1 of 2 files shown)


libraries/libxml/source/org/pentaho/reporting/libraries/xmlns/writer/XmlWriterSupport.java
```diff
@@ -923,63 +757,15 @@ public void writeTextNormalized( final Writer writer,
    * @param transformNewLine true, if a newline in the string should be converted into a character entity.
    * @return the normalised string.
    */
-  public static String normalize( final String s,
-                                  final boolean transformNewLine ) {
+  public String normalize( final String s, final boolean transformNewLine ) throws IOException {
```
tmorgner: Changing a method from static to non static is a breaking API change. It will break code that is compiled against pure 7.0 artifacts. That is a big NO NO in a minor release. 

Duplicating code is normally not a nice thing to do, but here you will have to do it.

tmorgner:
Yes, it is a hard requirement to support encodings other than UTF-8. That functionality is used in the XHTML export. Any character not found in the specified character set must be encoded using the unicode-escape entity (&uXXXX;).

AliaksandrDrebenchuk:
@tmorgner Previously was (&#xhhhh;) format like here https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references. But I have no idea how to read both of them from xml. Should I create custom handler? By the way, the biggest problem is this symbol http://unicode.org/cgi-bin/GetUnihanData.pl?codepoint=20bb7 which requires two chars instead one. It's a problem for many decoders.

tmorgner:
This is not about reading, you need to WRITE them. The numerical character references are an integral part of every standard conforming XML parser. So when you read an XML file with such a reference, the _parser_ takes care of the processing for you.

https://en.wikipedia.org/wiki/Numeric_character_reference

To make the XmlWriter work, you have to test each character written against the character-set given. Then, if the character is NOT included in the character set, you have write it as numerical character reference instead.

The problem with the old code was that it did this part incorrectly. Java uses UTF-16 as standard encoding for strings, so any extended character that is outside of that range (codepoint > 0xFFFF) is written as two characters from the so called "surrogate plane". When you encounter a character that is a surrogate, you cannot write it out normally. You have to first construct the real code-point for that character from the low- and high-surrogate character and then write that codepoint instead.

If you write a surrogate character as numeric character reference, any standard conforming parser will reject the result as surrogate characters are not a valid unicode character. 

The solution is simple: Iterate over the string like in the first answer on Stackoverflow:

http://stackoverflow.com/questions/1527856/how-can-i-iterate-through-the-unicode-codepoints-of-a-java-string

You can then test each character on whether it can be natively encoded (warning: CharactersetEncoder only accepts char for its quick test. You have to construct a one-codepoint string to test it properly:

        final Charset charset = Charset.forName( "UTF-8");
        final CharsetEncoder charsetEncoder = charset.newEncoder();
        final StringBuffer b = new StringBuffer();
        b.appendCodePoint(0x103a1);
        Assert.assertTrue(charsetEncoder.canEncode(b));

If the characterset accepts the character, you can write the code-point directly. Otherwise: (canencode returns false), you write the codepoint as numeric character reference. 

So basically: Do what the old code did, but dont use 16-bit "char" variables, use codepoints instead.

AliaksandrDrebenchuk:
@tmorgner You wrote very helpful and clear information. But my main problem is 
```java
    try (OutputStream out = new FileOutputStream( "c:\\out.txt" );
        OutputStreamWriter writer = new OutputStreamWriter( out, "cp1251" );) {
      writer.write( "\uD842\uDFB7" ); // write 𠮷
    }
    FileInputStream in = new FileInputStream( "c:\\out.txt" );
    InputStreamReader reader = new InputStreamReader( in, "cp1251" );
    String s = IOUtils.toString( reader );
    try (OutputStream out = new FileOutputStream( "c:\\out2.txt" );
        OutputStreamWriter writer = new OutputStreamWriter( out, "utf-8" );) {
      writer.write( s );
    }
```
Can you show me how to get '𠮷' into out2.txt? I tried write
\uD842\uDFB7
&uD842&uDFB7
&#xD842&#xDFB7
Nothing works
{code}
Can you show me how to get '𠮷' into out2.txt? I tried write
\uD842\uDFB7
&uD842&uDFB7
&#xD842&#xDFB7
Nothing works

tmorgner:
If you see characters in the range of D8xx and DFxx in your output then your code does not work correctly. D8 and DF are surrogate character ranges and therefore not valid XML characters.

Your test actually writes a extended Unicode character into a text file encoded with an ancient 8 bit encoding. There is nothing valid in that code fragment.

If you want to test how an XML parser handles extended characters, then you'll have to use an XML parser to actually interpret an XML stream. 

The code point you are trying to write is "0x20bb7", not that two-char sequence you actually write (and which is not representable in Cp1252). 

Try to write a XML fragment like this:
      
      <text>&#x20bb7;</text>

and parse it with an XML parser. You should get the correct result back with that:

    @Test
    public void testParse() throws Exception {
        // there is no easy way to produce text with extended code-points other than this here:
        String textOriginal = new String(new int[]{0x20bb7}, 0, 1);

       // note how this xml file contains the code-point, not the two UTF-16 encoded characters ...
        String xml = "<?xml version=\"1.0\" encoding=\"ASCII\"?>\n<text>&#x20bb7;</text>";
        final DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
        final DocumentBuilder db = dbf.newDocumentBuilder();
        Document d =  db.parse( new InputSource( new StringReader(xml)) );
        String text = d.getDocumentElement().getTextContent();
        Assert.assertEquals(text, textOriginal);
    }

tmorgner:
You have an import or OWASP encode, but (a) the OWASP jar is not part of LibXML because (b) the import is not used and should be removed.

AliaksandrDrebenchuk:
@tmorgner I know. Please don't worry before appear successful wingman report.

kolinus:
@tmorgner @pamval could you please review

PR #892, opened 2016-11-29
SP-3045 Chart Background not showing in Export to PDF - Insufficient data for an image

AliaksandrDrebenchuk:
image deflating removed


mchen-len-son:
Double-checked locally.
The integration tests failures look not related to the changes, and they are not reproducing locally.
Merging.

PR #893, opened 2016-11-30

YuryBY:
@mbatchelor, could you please take a look?

### DIFF
--- a/libraries/libbase/source/org/pentaho/reporting/libraries/base/util/XMLParserFactoryProducer.java
+++ b/libraries/libbase/source/org/pentaho/reporting/libraries/base/util/XMLParserFactoryProducer.java
@@ -1,0 +1,42 @@
+/*!
+* This program is free software; you can redistribute it and/or modify it under the
+* terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+* Foundation.
+*
+* You should have received a copy of the GNU Lesser General Public License along with this
+* program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+* or from the Free Software Foundation, Inc.,
+* 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*
+* This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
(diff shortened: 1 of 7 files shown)


libraries/libbase/source/org/pentaho/reporting/libraries/base/util/XMLParserFactoryProducer.java
```diff
@@ -0,0 +1,42 @@
...
+   * Enabling this feature prevents from some XXE attacks (e.g. XML bomb)
+   * See PPP-3506 for more details.
+   *
+   * @throws ParserConfigurationException if feature can't be enabled
+   *
+   */
+  public static DocumentBuilderFactory createSecureDocBuilderFactory() throws ParserConfigurationException {
+    DocumentBuilderFactory docBuilderFactory = DocumentBuilderFactory.newInstance();
+    docBuilderFactory.setFeature( XMLConstants.FEATURE_SECURE_PROCESSING, true );
+    docBuilderFactory.setFeature( "http://apache.org/xml/features/disallow-doctype-decl", true );
```
tmorgner: I would argue that this should be wrapped in a try-catch. If you choose a non-apache based XML parser, you should not end up with an unrecoverable error. Log a warning, and then let the user parse his XML file.

YuryBY: @tmorgner, you are absolutely right. .setFeature can throw ParserConfigurationException. That's' why this method throws exception in order to any client code has possibility process it (e.g. logging, try another parser or something else). Currently the same logic is implemented in other Pentaho projects:
https://github.com/pentaho/pentaho-kettle/blob/f7dfc8c65f7b37ae0e3fcdaa1a68c8f88a9cddbb/core/src/org/pentaho/di/core/xml/XMLParserFactoryProducer.java#L49
https://github.com/pentaho/pentaho-platform/blob/7dfbe9fd11e56a6df784589a4e4c5926b6b4a8ab/core/src/org/pentaho/platform/util/xml/XMLParserFactoryProducer.java#L50
https://github.com/pentaho/mondrian/blob/1712450ac2b25b778fedb729639816bb8941c652/mondrian/src/main/java/mondrian/util/XmlParserFactoryProducer.java#L45

PR #895, opened 2016-12-13
[BACKLOG-6746] - As a report designer, I want to use 'formulas' to set Global chart properties so these could be driven by programmatically control.

Aliaksandr-Kastenka:


Aliaksandr-Kastenka:
@tmorgner @EgorZhuk 
Please review.

tmorgner:
@wingman-pentaho Can you please re-order the ANT commands so that modules are built in the correct order? Building report-designer before building the engine-core or libraries is a fun exercise but not a useful one.

The build order should be "engine/core", "engine/legacy-charts", "report-designer" and finally "report-designer-charts".

Aliaksandr-Kastenka:
@wingman-pentaho

Aliaksandr-Kastenka:
@wingman-pentaho

tmorgner:
As Wingman remains badly configured, checked manually. Merged.

smaring:
call me crazy, but shouldn't things at least compile before being merged?

EgorZhuk:
@smaring the build on wingman has wrong order.  Libraries -  engine - designer is what we need here.

smaring:
trying to confirm by forcing the build order on devci

smaring:
compilation confirmed ... devci picked up the changes in the wrong order ... looks OK now

EgorZhuk:
@smaring thank you!

tmorgner:
@smaring I had to validate manually that it does compile and that it does not break things.

For reporting, Wingman is only good for very very simple bug-fixes - it breaks as soon as a change affects more than one module. Wingman builds modules in reverse order, and that never works. I tried to warn @wingman-pentaho months ago about the build order. When you look at the order of the ant-builds, you can tell it aint gonna work.

PR #896, opened 2016-12-13
Display HTML links within Report Designer elements

GregZuber:
PRD-5524


tmorgner:
Hi Greg. We cannot make changes to release branches. Branches that represent a specific release (7.0.0.1) normally do not receive fixes unless there is a breaking bug in it that must be fixed against that specific version. Maintenance branches for releases (ie 7.0) only receive bug-fixes to avoid breaking existing customer deployments. New features may alter how their reports behave - customers dont like surprises. 

New features like this always go into master and thus become a part of the official next release. Its ok to introduce new behaviours and features there, as this is what everyone expects from new major versions. Therefore: Please make this a pull request against master.

GregZuber:
Hi Thomas,

Thank you very much for your answer. I made a pull request to 7.0.0.1 as master is not compiling. I have an error:
Error:(75, 47) java: non-static method normalize(java.lang.String,boolean) cannot be referenced from a static context

But now I've just created the second pull request to master.

I work at CERN and it will be really useful for us if you can merge this feature. It will make our work much easier. So I'm closing this pull request, as I open the new one to master.

Thank you.

Best regards,
Grzegorz Zuber

PR #897, opened 2016-12-13
PRD-5776  Unable to open saved prpt report having Japanese surrogate characters

AliaksandrDrebenchuk:
XmlWriterSupport API was restored

### DIFF
--- a/libraries/libxml/source/org/pentaho/reporting/libraries/xmlns/writer/XmlWriterSupport.java
+++ b/libraries/libxml/source/org/pentaho/reporting/libraries/xmlns/writer/XmlWriterSupport.java
@@ -17,11 +17,6 @@
 
 package org.pentaho.reporting.libraries.xmlns.writer;
 
-import org.pentaho.reporting.libraries.base.util.FastStack;
-import org.pentaho.reporting.libraries.base.util.ObjectUtilities;
-import org.pentaho.reporting.libraries.base.util.StringUtils;
-import org.pentaho.reporting.libraries.xmlns.common.AttributeList;
-
 import java.io.IOException;
 import java.io.StringWriter;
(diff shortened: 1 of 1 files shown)


pamval:
the commit message is wrong, please amend.

AliaksandrDrebenchuk:
@pamval fixed

pamval:
@tmorgner, review?

libraries/libxml/source/org/pentaho/reporting/libraries/xmlns/writer/XmlWriterSupport.java
```diff
@@ -794,14 +803,19 @@ public void writeTextNormalized( final Writer writer, final String s, final bool
    * @param transformNewLine true, if a newline in the string should be converted into a character entity.
    * @return the normalised string.
    */
-  public String normalize( final String s, final boolean transformNewLine ) throws IOException {
+  public static String normalize( final String s, final boolean transformNewLine ) {
```
tmorgner: Please remove the "static". Within a point release you are not allowed to change the public API of existing classes. Making a method static is such a breaking change, and worst of all, it will break all jars that are compiled against older versions of the library.

AliaksandrDrebenchuk: It's a small urgent fix for previous PR. https://github.com/pentaho/pentaho-reporting/pull/889. The method "normalize" already was static.

pamval: Interestingly enough, there's a comment from Thomas on the previous commit warning that you should not change the accessibilty of the method

PR #900, opened 2016-12-15
[PRD-5716] Running a report that returns multiple RTF worksheets (Flo…

stanislau-strelchanka:
[PRD-5716] Running a report that returns multiple RTF worksheets (Flow Mode) runs into performance issues


stanislau-strelchanka:
@mkambol, @mdamour1976, @tmorgner, @mchen-len-son, @pamval please review

lucboudreau:
This PR was built twice in parallel on Wingman. I've canceled one of the executions, and the second will complete in a few minutes. Please do not schedule builds manually, especially PRD, which takes 90 minutes to build.

tmorgner:
Looks good. Lets wait for Wingman. Poor guy needs a bigger machine. I have an old C64 sitting around, if you want it. Maybe that will speed things up :)

@lucboudreau 90 mins sounds like head and base are built one after the other. Assuming there are spare CPUs, do you think it would be feasible to let both run in parallel? After all, they are independent builds until the very end when you merge the changes in test-failures and coverage.

tmorgner:
Looks good to me. Reported coverage changes on EHCache and the class-factory is the normal fluctuation we all got used to here.

PR #901, opened 2016-12-16
[PRD-5716] Running a report that returns multiple RTF worksheets (Flo…

stanislau-strelchanka:
[PRD-5716] Running a report that returns multiple RTF worksheets (Flow Mode) runs into performance issues - fix subsequent runs


stanislau-strelchanka:
@tmorgner please review

PR #904, opened 2016-12-29
PRD-5816 Cannot filter by null values in PRD and Report Viewer

AliaksandrDrebenchuk:
changing "= null" to "is NULL" was added

### DIFF
--- a/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/misc/datafactory/sql/SimpleSQLReportDataFactory.java
+++ b/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/misc/datafactory/sql/SimpleSQLReportDataFactory.java
@@ -12,23 +12,11 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2001 - 2016 Object Refinery Ltd, Pentaho Corporation and Contributors..  All rights reserved.
+ * Copyright (c) 2001 - 2017 Object Refinery Ltd, Pentaho Corporation and Contributors..  All rights reserved.
  */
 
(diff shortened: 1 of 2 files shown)


kolinus:
@tmorgner, @EgorZhuk   could you please take a look

EgorZhuk:
@tmorgner @kolinus  didn't work for me at least with inmemory sampledata report and custom postgres. Probably works with mysql

tmorgner:
No way I let this pass. We do not rewrite queries in the SQL datasource. We simply cannot guarantee that we do not break existing reports. JDBC drivers are widely varying in their support for features and the general contract around this is: The user is responsible for generating correct SQL.

JDBC-lite drivers like Hive or certain crude proprietary JDBC interfaces may not support the IS NULL and may error out now. Remember: Just because some databases support a feature does not mean all do. 

For null tests, a raw SQL user has to use the coalesce function, as explained in this forum thread.

http://forums.pentaho.com/showthread.php?74001-PRD-3-5-creation-of-SQL-query-for-conditional-null-or-non-null-value

To fix PIR (as in the case referenced in PRD-5816) implement this fix in the SQLGenerator of the metadata layer. At least there you have information about the SQL dialect used and can make a sane decision on whether your construct will be supported. And most importantly: You are not breaking other peoples reports.

Introducing a potentially breaking change like that will only pass my filters if it is guarded by a feature switch and if the original behaviour is the default.

tmorgner:
In addition: Your change unconditionally breaks queries that use parameter values in a function:

       WHERE aColumn = CONCAT("Prefix", ${param}, "suffix") 

would be rewritten to:

       WHERE aColumn = CONCAT("Prefix", ${param} IS NULL, "suffix") 


If you want to safely rewrite SQL, you have to be aware of the context where a variable is used. That requires a proper SQL parser and a lot more code than just a string replacement. You'd also would need to be aware of the underlying database (SQL has so many extensions that the term "standard" does no longer apply to it).

PR #910, opened 2017-01-18
[PRD-5236]-HTML ordered (<ol>) and unordered (<ul>) lists are not respected when rich-text-type = text/html

denisprotopopov:
- added rendering for <ol>,<ul>
 - added unit test


denisprotopopov:
@tmorgner please look at the changes

tmorgner:
Crude, but effective. Years ago, when writing the first version I looked at the OL and UL related styles rules and skipped it because they have some weird stuff going on that you can configure there. 

Getting at least a base line into it is a good choice. I probably should have done that at the time. So thank you for adding the missing bits here. And if later we want to support other bullet or numbering schemas, then at least your code makes this easy to add.

PR #911, opened 2017-01-20
[PRD-5563] - added flag -Dswing.useSystemFontSettings=false to report…

RPAraujo:
…-designer.bat

@tmorgner @pamval @pedrofvteixeira please review

According to documentation: http://docs.oracle.com/javase/7/docs/technotes/guides/swing/1.4/w2k_props.html
This can be a problem in font compatibility, adding this flag fixes the issue and only affects Windows clients.


tmorgner:
Sweet. A nice, non intrusive fix. I love it.

PR #914, opened 2017-01-25
[SP-3241][PRD-5236]-HTML ordered (<ol>) and unordered (<ul>) lists are not respected when rich-text-type = text/html

denisprotopopov:
- added rendering for <ol>,<ul>
 - added unit test


mchen-len-son:
This is weird, but those integration tests are passing locally, both under jacoco run and plain junit.

PR #916, opened 2017-01-27
[BACKLOG-14248] When starting PRD there is an error: Error opening bl…

andrei-viaryshka:
…ueprint xml url java.lang.NullPointerException

Exclude jars that cause problems.


andrei-viaryshka:
@smaring Could you review please?

PR #919, opened 2017-02-16
[SP-3259][PRD-3427]-The specified series-color is not applied to Radar Chart

denisprotopopov:
- now series-color applied to RadarChart
 - added unit test


denisprotopopov:

PR #926, opened 2017-03-02
[PDI-16039] - MIN() throws an exception instead of returning 0.

mbatchelor:


tmorgner:
Totally sane. Ignoring the errors, they smell like fallout or unresolved issues from the recent mavenization.

PR #927, opened 2017-03-07
[BACKLOG-15189]-Revert the changes to a XulDatatbaseDialog made durin…

EgorZhuk:
…g mavenization

Was commented during a mavenization and forgotten.


EgorZhuk:
@tmorgner ivy used to run tests in an alphabetical order, maven uses OS-dependent file ordering. Will create a suite for the failing test to eleminate it

tmorgner:
Order should not matter for that test. Something else is going on. Line 58 of that test class does not contain any code that dereferences a pointer. How can that code therefore fail at that position with an NPE?

I wonder whether creepy JaCoCo is failing instead ..

PR #930, opened 2017-03-21
[PRD-5878]-Remove element from a list, the right way

EgorZhuk:
That's why they teach us to use iterators, when removing elements from a collection...

What happens here - we remove an element from a list and decrement it's size.
The size is not stored in a variable and is retrieved each time.
As a result we have only every other row in the report.

Simple demo: https://gist.github.com/EgorZhuk/34950613e64424c2b75ed13bbe137183


tmorgner:
Bah, you and your modern approaches. In my time we haven't had iterators, we had to manually herd electrons through the transistors each time ... 

Oh, and totally approved.

PR #939, opened 2017-04-03
[PDR-5563]

RPAraujo:
added - Dswing.useSystemFontSettings=false to report-designer.bat this will keep java from changing the font settings to System


RPAraujo:
@pamval @mbatchelor please review
made a new pull request because the old one somehow got lost.

tmorgner:
Seems safe enough. Merged.

PR #940, opened 2017-04-03
Revert "[PRD-3638]-Check the platform credentials faster"

EgorZhuk:
This reverts commit 276e6122bdd3f1119acc8d6ae92964d6871b4fb3.

No difference in a loading speed - seems PRD-5300 already fixed covered the fix.


tmorgner:
Bugger! I would expect that fix only to make a difference if you have a hell of a lot of files, as for steelwheels the server is always fast.  

Next thing to try would be to change VFS to load the tree in stages. Instead of using infinite depth, use depth of 2 or so, and mark the leaf-directory nodes as "dirty" so that they get reloaded if and as needed.

PR #944, opened 2017-04-12
[PRD-5873] Underlined Label Elements in PRPT Report Generate Adobe Acrobat Error When Executed in PDF

Advent51:
- the order of drawing graphics and text closing was changed: now we firstly close the text block and then draw a graphics.
- test was added


Advent51:
@tmorgner please review

PR #947, opened 2017-04-12
[PPP-3709] - Upgrade POI to 3.15

mbatchelor:


engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/modules/output/table/xls/helper/HSSFCellStyleProducer.java
```diff
@@ -256,7 +256,7 @@ private static Color createColor( final XSSFColor color ) {
       if ( color == null ) {
         return null;
       }
-      final byte[] rgb = color.getRgb();
+      final byte[] rgb = color.getRGB();
```
tmorgner: I love the smell of subtle API changes in the morning. :)

tmorgner:
@mbatchelor I would assume this merge problem is a @wingman-pentaho error. Either the build process for wingman actually edits the POM file, or git on that node is misconfigured.

@wingman-pentaho Would a crude "git reset --hard" before you merge solve the problem? After all, any part of the build or unit-test process could have produced or touched some files that then prevents a merge (independent of whether *your* process does it today, we had hsql file changes in the past doing exactly that ). A "git reset" would brutally and nicely revert such changes.

PR #948, opened 2017-04-12
[BACKLOG-15692]-Fix order dependent tests

EgorZhuk:
Fixed naughty tests and removed the ordering from the POM.
Checked with a 'reversealphabetical' setting and a default one on linux and windows (the default is a 'filesystem' and it is OS dependent).


tmorgner:
Regarding the AttributeMetaRegistryTest: That test checks the registration and lookup process.  The reflection makes the test awful. :) 

Wouldnt it be better to make ElementTypeRegistry's constructor package-protected, then work the test from a local instance? Package protected is "good enough" to prevent silly users to mess with the system, but it cuts out a lot of the reflection hacks you have to go through. And it guarantees that you never pollute the global instance with test data.

    final ElementTypeRegistry registry = new ElementTypeRegistry();
    registry.registerFromXml( AttributeMetaRegistryTest.class.getClassLoader().getResource
        ( "org/pentaho/reporting/engine/classic/core/metadata/meta-elements-data.xml"  ));

and then change all `ElementTypeRegistry.getInstance()` in that test to just `registry`.

EgorZhuk:
Ok, will rework it.

EgorZhuk:
@tmorgner reworked, but seems poor wingman can not swallow the pom

EgorZhuk:
@pminutillo @hbfernandes any reasons why wingman blows up on a POM change?

hbfernandes:
I do not know, tried to understand why but with no luck so far. Can you check if the pom file has LF line endings and not CRLF? This has been known to cause this type of issue in the past.

EgorZhuk:
@hbfernandes the ending is UNIX style. Let's try an empty line in the end of file - it caused issues before.

tmorgner:
Egor, code looks sane and good. Seeing that Wingman throws such a party, lets see whether the line-ending is to blame. 

Otherwise we can continue blame Wingman directly, as usual. If we blame him often enough, he may turn from being a "build terminator" to a "general purpose terminator". When (not if!) that happens: I have a list of people to 'merge', just need a time machine and we can be back in the 80's.

EgorZhuk:
@hbfernandes @tmorgner the empty line didn't help, and as I've already said, the line ending of the file is LF

EgorZhuk:
@hbfernandes @tmorgner the build was aborted on the CI, and as far as I know we won't have any snapshots till the weekend, hence will need to wait for a QAT. Without a snapshot, we only can hope that a QAT build won't be broken.

EgorZhuk:
Nevermind, the changes in a schedule is for 6.1 and 7.0. I've restarted the build.

tmorgner:
Well, the problem Wingman has is a merge problem. Github is smarter than Wingman (no surprise, my broom is smarter than Wingman!) and has no merge problem. So CI should be fine, given that it does not try to use a broken merge strategy.

tmorgner:
Damn, my broom just told me my floor coverage is down, even though I only updated the calendar on my wall. My broom turned into wingman, help! .

EgorZhuk:
@tmorgner your jokes will definitely be on the list of things I will miss when leaving the team

PR #956, opened 2017-04-26
[PIR-1338]-Interactive report with 4 or more columns to group gives wrong total

denisprotopopov:
- added set up groupname for correct calculating
 - added unit test


denisprotopopov:

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/output/AbstractReportProcessor.java
```diff
@@ -423,12 +425,22 @@ protected ProcessState processPrepareLevels( ProcessState state, final int maxRo
 
   protected abstract OutputFunction createLayoutManager();
 
+  void processGroups( MasterReport report ) {
```
tmorgner: No. You do not rewrite reports in the processor, you make sure that reports are defined correctly in the first place.

If PIR wants to reference a group, it should set its name inside the PIR code that generates the MasterReport. Making changes like yours here is totally breaking all oter reports that have named groups.

PR #958, opened 2017-04-27
[PRD-5882]-When double click in a table of contents shows an error message

denisprotopopov:
- simple element shouldn't be convert into complex element such as TableOfContent

### DIFF
--- a/designer/report-designer/src/main/java/org/pentaho/reporting/designer/core/actions/elements/MorphAction.java
+++ b/designer/report-designer/src/main/java/org/pentaho/reporting/designer/core/actions/elements/MorphAction.java
@@ -12,7 +12,7 @@
 * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Lesser General Public License for more details.
 *
-* Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+* Copyright (c) 2002-2017 Pentaho Corporation..  All rights reserved.
 */
 
 package org.pentaho.reporting.designer.core.actions.elements;
@@ -40,10 +40,10 @@ import org.pentaho.reporting.engine.classic.core.metadata.ElementMetaData;
 import org.pentaho.reporting.engine.classic.core.metadata.ElementType;
 import org.pentaho.reporting.engine.classic.core.util.InstanceID;
 import org.pentaho.reporting.libraries.base.util.ObjectUtilities;
-
-import javax.swing.*;
-import java.awt.*;
 import java.awt.event.ActionEvent;
+import java.awt.Image;
+import javax.swing.Action;
+import javax.swing.ImageIcon;
 import java.beans.BeanInfo;
 import java.text.DecimalFormat;
 import java.text.Format;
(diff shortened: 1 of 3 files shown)


designer/report-designer/src/main/java/org/pentaho/reporting/designer/core/actions/elements/MorphAction.java
```diff
@@ -106,6 +112,10 @@ public void actionPerformed( final ActionEvent e ) {
       if ( ObjectUtilities.equal( visualElement.getElementTypeName(), targetElementType.getMetaData().getName() ) ) {
         continue;
       }
+      if ( COMPLEX_ELEMENT_SET.contains( targetElementType.getMetaData().getName() )
```
mbatchelor: Shouldn't this logic be in the construction of the Morph menu ? It looks to me like they will still be able to choose to morph the field into a TOC, but then this code will just skip doing it. So the next bug will be "I choose morph to a TOC, and nothing happens." I believe the right place for this is in the menu construction. If that's not possible for whatever reason, then we have to pop a message to them telling them that the action doesn't make sense / cannot be done.

designer/report-designer/src/main/java/org/pentaho/reporting/designer/core/actions/elements/MorphAction.java
```diff
@@ -63,6 +66,9 @@
   private final ElementMetaData metaData;
   private ElementType targetElementType;
 
+  public static final String[] COMPLEX_ELEMENT_NAMES = new String[] { "band", "sub-report", "crosstab-report", "toc", "index" };
```
tmorgner: Use the metadata! The ElementType#getTypeClassification tells you whether a element is a Subreport or a control-object. (Control objects are things like Groups, which cannot be morphed either.

designer/report-designer/src/main/java/org/pentaho/reporting/designer/core/actions/elements/MorphAction.java
denisprotopopov: @tmorgner does It should works as Marc suggested?

tmorgner: Morphing has two inputs: What element do I start with, and what element type do I end up with. 

The menu choice gives you the target (ie: "Morph the selected element into a .. (label, text-field, etc)"). But the source of your morph is the current selection, and that is changing all the time. 

For target types that are eternally disabled, like the complex elements like subreports and groups, we do not generate the menu in the first place. We still have to handle the case where a user selects one of those elements as its source element and expects that to convert to something else.

denisprotopopov:

designer/report-designer/src/main/java/org/pentaho/reporting/designer/core/actions/elements/MorphAction.java
```diff
@@ -106,6 +106,10 @@ public void actionPerformed( final ActionEvent e ) {
       if ( ObjectUtilities.equal( visualElement.getElementTypeName(), targetElementType.getMetaData().getName() ) ) {
         continue;
       }
+      if ( targetMetaData.getReportElementType().compareTo( ElementMetaData.TypeClassification.SUBREPORT ) == 0
```
tmorgner: The lords of Java in their infinite wisdom invented Equals to test for object equality. Use it, they will be grateful!

PR #959, opened 2017-04-27
[PRD-5886] View Menu - The Grid Settings dialog fails to open because of java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Integer

Advent51:
- Type was incorrect - fixed.


Advent51:
Fix was tested on Windows and Ubuntu.

@mchen-len-son @mbatchelor @pamval please review

PR #963, opened 2017-05-23
[SP-3504] Backport of PRD-5886 - View Menu - The Grid Settings dialog…

Advent51:
… fails to open because of java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Integer (6.1 Suite)


Advent51:
@mbatchelor please review

Advent51:
@mchen-len-son please review

PR #970, opened 2017-06-01
[BISERVER-12968]-Self Links defined in a PRPT chart do not working on…

PashkovichAndrei:
… the

server

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/formula/EnvFunction.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/formula/EnvFunction.java
@@ -12,13 +12,16 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+ * Copyright (c) 2002-2017 Pentaho Corporation..  All rights reserved.
  */
 
 package org.pentaho.reporting.engine.classic.core.function.formula;
 
 import org.pentaho.reporting.engine.classic.core.ReportEnvironment;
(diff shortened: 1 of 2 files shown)


PashkovichAndrei:

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/formula/EnvFunction.java
```diff
@@ -47,6 +54,44 @@ public TypeValuePair evaluate( final FormulaContext context, final ParameterCall
 
     final ReportFormulaContext rfc = (ReportFormulaContext) context;
     final ReportEnvironment reportEnvironment = rfc.getProcessingContext().getEnvironment();
-    return new TypeValuePair( AnyType.TYPE, reportEnvironment.getEnvironmentProperty( fieldName ) );
+    Object property = reportEnvironment.getEnvironmentProperty( fieldName );
```
tmorgner: Nope, that is a bad hack. This introduces a hidden dependency on an implementation detail (ie how ReportFormulaContext is implemented) and makes the ENV function behave different from querying the report-environment outside of that function. When your problem is "How can I get a new property from the ReportEnvironment" I would think the most straight-forward solution is to ask "How can I make ReportEnvironment" return a value for that new property.

And what is "/viewer" doing in that code? That is a implementation detail of the report-viewer and has absolutely no place in the non-reportviewer codebase. 

A report that runs in the standalone engine here has actually no concept of a linkable self-url. Would you really expect a generated HTML or PDF document to link to a local file URL? And how is this "magic" of your's going to work when a report is not stored in the normal JCR, like a report produced by PIR, for instance?

You are implementing this on the wrong level. If the report-environment has no notion of a self-URL, let it return "null". That will make the DRILLDOWN formula return NULL as well - and that disables linking while the report is running outside of a server. You dont need any code for this behaviour. 

So close this Pull request and lets move on to the BI-server plugin - and specifically the PentahoReportEnvironment class. *That* is where the code for generating a "selfURL" response should live.

PR #974, opened 2017-06-06
[PRD-5631] If report-designer.sh is launched from any directory other than its own, karaf fails to start

Advent51:
- the problem with launcher directory was fixed


Advent51:
@mbatchelor @pamval please review

PR #975, opened 2017-06-16
[BISERVER-12968]-SelfLinks defined in a PRPT chart do not work on the

PashkovichAndrei:
server


PashkovichAndrei:

pamval:
@tmorgner, please review when you get the chance

PashkovichAndrei:
@tmorgner, @pamval, @mbatchelor, could you please do review?

PashkovichAndrei:

tmorgner:
No, No, No and double no. What is it with your tendency to wrap tentacles of unrelated code to a lowest level as possible? You are pretty much coding a manual formula evaluation into that place for no particularly sane reason. STOP DOING THAT!

The formula syntax is perfectly able to express that logic as a pure formula. So if you are unhappy with the formula that is executed, wouldn't it make much more sense to change the formula, than to hack a custom interpretation into the most random location of the formula interpreter? 

I close this pull request and beg you to try to be serious in your attempts to fix the bug.

PR #978, opened 2017-06-22
[SP-3637][PRD-5896]-Performance logger REPORT_QUERY is not properly initialized

stepanovdg:


stepanovdg:
@mbatchelor  @mchen-len-son Could you please take a look?

PR #982, opened 2017-06-26
[PRD-3023]-Named color code for elements is NOT the same for chart.

denisprotopopov:
- fixed. added unit tests


tmorgner:
Absolutely frigging not!. You are globally changing the green constant for all colour uses, both in charts, elements, and everywhere someone refers to colours by name. That is absolutely NOT COOL. Remove that so called fix immediately.

If you want to fix a isolated behaviour in charts, then how about actually looking at that charting subsystem, and finding out what the heck goes wrong there. Read through AbstractChartExpression and try to understand how that beast computes colours from strings, and then that code there.

PR #985, opened 2017-07-05
[PPP-3735] - Use of vulnerable component httpclient-3.0.1.jar, common…

YuryBY:
…s-httpclient-3.1 - httpclient-4.0.1.jar, SONATYPE-2007-0004, CVE-2012-6153, CVE-2011-1498, CVE-2014-3577

@mbatchelor, @pamval, could you please take a look?

### DIFF
--- a/libraries/libpensol/pom.xml
+++ b/libraries/libpensol/pom.xml
@@ -28,13 +28,19 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
+      <groupId>org.pentaho.reporting.engine</groupId>
+      <artifactId>classic-core</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
       <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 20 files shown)


YuryBY:
@mbatchelor, @pamval, @smaring, @tmorgner could you please take a look?

libraries/libpensol/pom.xml
```diff
@@ -118,6 +118,11 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
+      <groupId>pentaho-kettle</groupId>
```
pamval: @YuryBY again, bringing in kettle-core as a dependency for libpensol is to be avoided

tmorgner:
@pamval is this safe to merge or are there other dependencies in other projects that need to be merged first? The code itself here looks sane and good enough for a merge.

YuryBY:
Here is rebased PR: https://github.com/pentaho/pentaho-reporting/pull/991

PR #987, opened 2017-07-17
[PDI-14764] Jaro algorithm glitch

stanislau-strelchanka:


tmorgner:
seems safe to remove. Nothing in reporting relies on this library directly. If some other code (kettle) requires this functionality, they will pull it in via their dependencies.

PR #988, opened 2017-07-17
[PRD-3383]-Parameter Dialog on PRD-exports has no scrollbars, prevents

PashkovichAndrei:
users from entering parameters when long parameter list


PashkovichAndrei:
@tmorgner, could you please have a look?

tmorgner:
Trivial patch changing only non-public details. Merged.

PR #990, opened 2017-07-31
[PRD-3322]-On Linux environments, loading dialog doesn't go away when

PashkovichAndrei:
previewing a parameterized report without a query


PashkovichAndrei:

PR #992, opened 2017-08-01
[PRD-5914]-Property for "Word-Break" Unable to be Saved as "False" in Standard Report

denisprotopopov:
- fixed, added unit tests


denisprotopopov:

PR #993, opened 2017-08-02
[PRD-3954] DRILLDOWN Formula Editor, the Location resets unexpectedly

Advent51:
- removed unnecessary parameters filtration;
- added a check before event creation;
- fixed formula creation - created formula now is valid by default;
- fixed unreported bug with '=' duplication on location changing;


Advent51:
@pamval @mbatchelor @stepanovdg please review

kolinus:
@tmorgner could you please take a look

Advent51:
@pamval @stepanovdg @tmorgner please review

tmorgner:
This just removes some 4.8 area legacy code and thus prevents the bug. Anything that makes the code simpler is good.

PR #994, opened 2017-08-03
[CLEANUP] Fix dependency on commons-httpclient

DFieldFL:


kurtwalker:
wingman failed from nexus upgrade.  this should fix a compile error so I'll merge

PR #995, opened 2017-08-04
[PPP-3566] Use of vulnerable component xerces:xercesImpl:2.9.1and 2.11.0 - CVE-2009-2625 and CVE-2013-4002

stepanovdg:
- removed explicit xerces dependencies. test usages of xerces are only left


tmorgner:
Ahh, Xerces, the bane of my existence. Good to see that go. Should not have any impact on anything, as sane XML code does reference the parser indirectly via the JDK API. Not so sane dependencies should get Xerces (if they really want it) via their own transitive dependencies. A happy merge this is.

PR #999, opened 2017-08-17
[SP-3725][PRD-5914]-Property for "Word-Break" Unable to be Saved as "False" in Standard Report

denisprotopopov:
- fixed, added unit tests


tmorgner:
@denisprotopopov Wingman still complains about 2 formatting problems. If you think this is already fixed, please give Wingman a nudge to revalidate this PR.

PR #1003, opened 2017-08-23
[BACKLOG-17345] Reset system properties to null instead of empty string to avoid UnknownHostException with new Postgres driver

niehusa:
@pentaho/rebelalliance @pentaho/rogueone @tmorgner 
Please review.
This fixed 'UnknownHostException: localhost' exception thrown from the new driver.


tmorgner:
@niehusa There are code style violations. Other than that the PR is fine.

Oh, and ignore Wingman's coverage nagging. The coverage numbers are obviously generated by sending 1000 Ewoks though an electric storm and counting the hairs that fall out. In any case, Wingman is useless for that. If Emperor Palpatine were still alive such stuff would not happen!

niehusa:
@tmorgner Thank you so much for reviewing.
Ran checkstyle and added space accordingly. W/R/T to Wingman, I can't agree more :-P

PR #1017, opened 2017-09-22
[PRD-5916] To see the default date in the report it was necessary to …

stanislau-strelchanka:
[PRD-5916] To see the default date in the report it was necessary to open the parameter dialog two times (click OK)

### DIFF
--- a/designer/report-designer/src/main/java/org/pentaho/reporting/designer/core/util/table/ElementMetaDataTable.java
+++ b/designer/report-designer/src/main/java/org/pentaho/reporting/designer/core/util/table/ElementMetaDataTable.java
@@ -12,7 +12,7 @@
 * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Lesser General Public License for more details.
 *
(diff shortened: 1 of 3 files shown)


stanislau-strelchanka:
looks like a wingman issue as same tests are failing on an empty pr - #1019

@tmorgner @pamval @mchen-len-son please review

tmorgner:
Either that or Mondrian now as a new default behaviour and therefore the golden sample of these tests no longer applies. If that is the case and the new behaviour is not a regression, then we will have to regenerate the golden samples.

stanislau-strelchanka:
default behaviour was changed in mondrian due to MONDRIAN-1967.
created pr that updates expected results: #1020

@tmorgner @pamval @mchen-len-son please review

stanislau-strelchanka:
@tmorgner @pamval @mchen-len-son please review

stanislau-strelchanka:
@tmorgner @pamval @mchen-len-son please review

stanislau-strelchanka:
@tmorgner @pamval @mchen-len-son please review

tmorgner:
Looks good. Love the classic util-Date vs sql-Date bug fixed here.

PR #1018, opened 2017-09-25
[PPP-3735] - Use of vulnerable component httpclient-3.0.1.jar, common…

YuryBY:
…s-httpclient-3.1 - httpclient-4.0.1.jar, SONATYPE-2007-0004, CVE-2012-6153, CVE-2011-1498, CVE-2014-3577

@tmorgner, @duarteteixeira, @pamval, rebased PR. There are now compilation errors locally. Thanks.

### DIFF
--- a/assemblies/mac/src/main/resources/resource/resources/log4j.xml
+++ b/assemblies/mac/src/main/resources/resource/resources/log4j.xml
@@ -102,11 +102,11 @@
     <priority value="WARN"/>
   </category>
 
-  <category name="org.apache.commons.logging.simplelog.log.org.apache.commons.httpclient">
+  <category name="org.apache.commons.logging.simplelog.log.org.apache.http">  
     <priority value="WARN"/>
   </category>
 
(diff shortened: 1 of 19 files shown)


pamval:
@YuryBY @tmorgner what are these failing tests, can someone take a look please?

tmorgner:
The mondrian default behaviour seems to have changed in a recent commit and Mondrian now seems to report (at minimum, tests fail at first change) the visible property as true. 

If that is an expected change, then we have to regenerate the golden samples. If it is not an expected change, then Mondrian changes its returned values randomly and should be called Picasso.

pamval:
@YuryBY can you validate that Thomas is correct and that Mondrian's behavior has changed? 

If this is accurate, we need to figure out why it changed. I don't recall any changes to Mondrian that may have caused this.

pamval:
@YuryBY can you take a look and figure out if this is a result of the xstream upgrade in mondrian?
If mondrian has changed its behavior we need to understand when, how and why.

YuryBY:
@pamval, here the PR that should fix the tests: https://github.com/pentaho/pentaho-reporting/pull/1020

PR #1026, opened 2017-10-04
[BACKLOG-19148]-java.lang.UnsupportedOperationException error in Pentaho Report Designer when publishing report to the server

stepanovdg:
Removed org.apache.karaf.jaas.modules-x.x.x.jar http://jira.pentaho.com/browse/BACKLOG-18473 
Contained old version of http-client inside.

### DIFF
--- a/assemblies/pom.xml
+++ b/assemblies/pom.xml
@@ -184,11 +184,6 @@
     </dependency>
     <dependency>
       <groupId>org.apache.karaf.jaas</groupId>
-      <artifactId>org.apache.karaf.jaas.modules</artifactId>
-      <scope>compile</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.karaf.jaas</groupId>
       <artifactId>org.apache.karaf.jaas.boot</artifactId>
       <scope>compile</scope>
     </dependency>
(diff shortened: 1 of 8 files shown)


stepanovdg:
@tmorgner Could you please take a look?

stepanovdg:
@tmorgner @duarteteixeira @pamval @mchen-len-son @kolinus Could you please take a look?

stepanovdg:
@tmorgner @duarteteixeira @pamval @mchen-len-son @kolinus Could you please take a look?

tmorgner:
You are handling circular redirects in the VFS handler. If I remember that right, stuff like that was/is caused by the Spring Security intercept that checks authentication credentials (if no cookie is set). That would indicate that both CDA and the report-designer module need the same config. These targets also hit the same Pentaho server with the very same Spring security playing silly games.

stepanovdg:
@tmorgner  I just update the way params are set into client request. 
So if previously was asked for circular redirects it proceed doing that. 
We have a regression related to httpclient update. Currently should use the new way of setting params.

stepanovdg:
@tmorgner @duarteteixeira @pamval @mchen-len-son @kolinus Could you please take a look?

tmorgner:
Well, it seems to do same stuff as before, so it should at least not be worse than before.

PR #1028, opened 2017-10-04
[BACKLOG-19185] - fix NPE when exporting to RTF format

candentira:


candentira:
@tmorgner Could you please review?

PR #1031, opened 2017-10-12
[PRD-5634] PRPT with PDI KTR datasource where Table Input step takes …

stanislau-strelchanka:
[PRD-5634] PRPT with PDI KTR datasource where Table Input step takes parameters throws error in PRD


stanislau-strelchanka:
@tmorgner please review

PR #1037, opened 2017-10-20
[PRD-5925] In PRD, opening a file containing surrogate pair character…

stanislau-strelchanka:
[PRD-5925] In PRD, opening a file containing surrogate pair characters in the file name from the repository results in an error.


stanislau-strelchanka:
For publishing a report default java encoder was used. For retrieving it from repository custom encoder
was used. Custom encoder is having problems with some of non standard chars. Replaced it with java in libraries/libpensol module.

@tmorgner please review

stanislau-strelchanka:
@tmorgner please review

tmorgner:
Please deprecate the broken URLEncoder in Libbase (with a reference to this case) and also fix the other use of this encoder in RepositoryPublishDialog. Otherwise you'd only fix half the occurrences, and publishing a file with characters from the supplementary planes would still be broken. So you could open the file, but never write it back to the server.

stanislau-strelchanka:
@tmorgner updated, please review

tmorgner:
Perfect. Merged.

PR #1052, opened 2017-10-27

stanislau-strelchanka:
[PRD-4595] Errors from the datasource used in a SINGLEVALUEQUERY function are consumed silently

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/FormulaExpression.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/FormulaExpression.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2001 - 2013 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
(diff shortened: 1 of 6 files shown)


stanislau-strelchanka:
Found 'failOnError' property in FormulaExpression. Couldn't found any usage of it in code(except for IT). It is always null and assumed to be false. Added 'failOnError' method for function interface in the current pr.

@tmorgner can you please review it?

tmorgner:
I disagree with this solution. You will create "magic" behaviour with that. 

The fail-on-error flag is a property of the formula evaulation and tells us how to handle any error that bubble up to the formula level. All kinds of formula can throw errors, including calculations on null-values, division by zero, failed lookups and so on. "fail-on-error" defines a general error handling policy for that formula function instance, 

Users can set the "fail-on-error" property in the report-designer UI. This is an expert-level property, so you only see it if you have the corresponding option enabled in the report-designer preferences. Properties are set via reflection, so you wont necessarily find code references to it. By making this a property, users can choose the error handling strategy based on the formula's use case. We do not necessarily want to fail a report on all errors, only on critical ones. Forcing users to hunt down every possible error state is not nice.

Basically: If "fail-on-error" is set to true, and the formula evaluates with any error (DIV/0, NA, or a error triggered by the query function, then the report should abort. 

If the error is caught by the formula (ISERROR(..)) function, then the report should not fail.

If "fail-on-error" is set to false, the report should never fail, regardless what the status of the formula evaluation is. 


Note: This is a very old bug-report, so the functionality may already be in the product and just needs revalidation. We had plenty of duplicate cases in the past and not all duplicates got always tracked down and closed in a reliable fashion.

PR #1061, opened 2017-11-01

stanislau-strelchanka:
[PRD-4595] Errors from the datasource used in a SINGLEVALUEQUERY function are consumed silently

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/FormulaFunction.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/FormulaFunction.java
@@ -12,13 +12,14 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2001 - 2013 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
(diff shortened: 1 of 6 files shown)


mchen-len-son:

mchen-len-son:

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/FormulaFunction.java
```diff
@@ -83,7 +84,9 @@ public FormulaFunction() {
   }
 
   public Boolean getFailOnError() {
-    return failOnError;
+    return failOnError == null
```
tmorgner: We use reflection to read and write properties for expressions. Hijacking the getter to never return null means the UI never sees null, and as soon as a user sets the value to null, your default catch comes in and "lies" that the value is whatever the system config says.

Likewise when saving the report: We use reflection for that as well, and call this getter. So we never write null anymore, and upon parsing the PRPT there will be a defined value here. So a loaded report will never fall back to the config. 

Work around that by introducing a new getter-only property "getComputedFailOnError" or so, that returns true or false based on your fallback logic. This way the original getter does not have to "lie" and both the expression's logic and the xml-writer and GUI-editors can go on living happily ever after.

mchen-len-son:
thanks @tmorgner.
updated the changes and issued a new PR #1067.

closing this one.

PR #1067, opened 2017-11-14

mchen-len-son:
…tion are consumed silently


mchen-len-son:

tmorgner:
This looks perfect to me. Merged.

PR #1068, opened 2017-11-16
[PRD-5925] In PRD, opening a file containing surrogate pair characters in the file name from the repository results in an error.

stepanovdg:


stepanovdg:
@tmorgner @pamval  Could you please take a look? That is fix for regression caused by previous commit.

mchen-len-son:
This was tested and validated by QA on 7.1.
Considering the code is pretty same, merging.

PR #1069, opened 2017-11-16
[SP-3851][PRD-5925] In PRD, opening a file containing surrogate pair characters in the file name from the repository results in an error.

stepanovdg:


stepanovdg:
@mchen-len-son Could you please take a look?

PR #1072, opened 2017-11-28
[PRD-5932] When using 'Row' layout in PRD, report elements set to 'vi…

pentaho-lmartins:
…sible=false' still display on Excel or Excel 2007 output

@duarteteixeira @pamval @mbatchelor

### DIFF
--- a/engine/core/pom.xml
+++ b/engine/core/pom.xml
@@ -5,10 +5,10 @@
   <parent>
     <groupId>org.pentaho.reporting.engine</groupId>
     <artifactId>reporting-engine</artifactId>
-    <version>8.1-SNAPSHOT</version>
+    <version>8.1.0.0-SNAPSHOT</version>
   </parent>
   <artifactId>classic-core</artifactId>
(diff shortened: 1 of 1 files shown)


mchen-len-son:
@tmorgner please take a look.

tmorgner:
One thing: You cannot use "Element#isVisible" in here. That style collection contains only half the possible values and would be wrong if you use stylesheets. Use "Element#getComputedStyle#getStyleProperty" instead. This returns the computed value after the style resolver has done its magic.

mchen-len-son:
@pentaho-lmartins can you please update changes considering the recommendations from Thomas?
thanks.

pentaho-lmartins:
Reviewed previous pull request and added unit tests.
@mchen-len-son @tmorgner @pamval @duarteteixeira

mchen-len-son:
@pentaho-lmartins can you please squash the commits?

pentaho-lmartins:
Closing this PR. Will add a new one shortly.

PR #1075, opened 2017-12-02

mchen-len-son:
…tion are consumed silently

 - logging error when SingleValueQuery cannot be evaluated


mchen-len-son:

When opening that broken report in Pentaho Server, the root error is not logged on server's console, and you cannot see what's wrong there. The evaluation exception, which is thrown after, just says "unexpected" (see [code](https://github.com/pentaho/pentaho-reporting/blob/0a95fff61c3b37ae7bb19cd4cd8020c23af4cf16/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/formula/SingleValueQueryFunction.java#L124)).
Not sure if this is the right place to fix.

mchen-len-son:

PR #1082, opened 2018-01-09
[BACKLOG-12466] Add KettleVFS support for reporting

AlexanderBuloichik:


tmorgner:
This does not work. Kettle Core depends on Reporting-Engine-Core, and now Engine-Core depends on Kettle-Core creating a circular dependency. 

Given that your problem lies within Kettle, I would recommend to fix this in Kettle's reporting step instead. Pentaho Reporting provides methods that allow you to write data to streams and the HTML output can be configured (in the client code, outside of the engine-code) to write to the JCR. 

The reporting engine uses the interfaces in LibRepository for this purpose. Please do have a look at the reporting-plugin of the BI-Server, especially the HTML-to-JCR output we have there. It is practically a blueprint for what you want to achieve within Kettle's reporting step.

AlexanderBuloichik:
@tmorgner Thomas, kettle-core doesn't depend on reporting-core. Probably you mean kettle-engine. I don't see circular dependencies.
Implementation of LibRepository interface will not be enough because PdfExportTask should be changed to work with KettleVFS.

tmorgner:
OK, lets spell this out: Kettle will be a *client* to a reporting output that is perfectly able to work with byte-streams. Why would you need to change the PDF code here if all you want to do is call a single [method](https://javadoc.pentaho.com/reporting/pentaho-reporting-engine-classic-core/org/pentaho/reporting/engine/classic/core/modules/output/pageable/pdf/PdfReportUtil.html#createPDF(org.pentaho.reporting.engine.classic.core.MasterReport,%20java.io.OutputStream)) that writes its data into an output stream.

Tell me, why would this class even be aware of Kettle's VFS stuff? It does not need to, your ReportingStep should simply call a different method, and pass the VFS's output stream to that method I quoted above. 

Same for RTF, Excel, CSV and all other Single-File outputs. 

As I said: HTML is slightly different. But you already did the sensible thing and implemented a VFS handler for LibRepository, Now all you need to do is to keep the Kettle specific stuff inside Kettle (ie the bit where you resolve a string into a VFS-FileObject) and simply allow your HTML export to use your VFS code.

If you are hesitant to copy the HtmlStreamExportTask.java into kettle, then that is fine. Simply change the class here to allow you to provide your own repository from the given string using an protected method

    protected Repository createRepositoryFromPath (String path)

This way you can subclass that HtmlStreamExportTask in your Kettle code and use Kettle's VFS magic to resolve the path without introducing a dependency on Kettle here.

PS: Do not deprecate the java.io.File stuff. File is NOT the same as VFS-File, there are slight variations on how stuff is handled internally and unless you want to make a formal proof that VFS behaves exactly as a local file system in all circumstances (hint: It does not on Windows or when dealing with locking and race conditions around temp-file creation) you better do not touch the existing code. A java.io.File is NOT the same as a VFS-FileObject and even Apache-VFS would not claim that.

AlexanderBuloichik:
@tmorgner Okay Thomas, I hope I understand. You means that I shouldn't use PdfExportTask(and other tasks), but create own implementation or extend these tasks for ability to use VFS. Will try.

PR #1089, opened 2018-01-25
[BACKLOG-20807] Added visibility to MQTT jar to OSGI

DFieldFL:


DFieldFL:
@pentaho/ackbar please review

PR #1091, opened 2018-01-29
[SP-4053] Backport of PRD-5932 - When Using 'Row' Layout in PRD, Repo…

pentaho-lmartins:
…rt Elements Set to 'visible=false' Still Display on Excel or Excel 2007 Output (7.1 Suite)

@pamval @duarteteixeira


pentaho-lmartins:
@pamval Added test cases from master to this branch. Squashed both commits.

PR #1092, opened 2018-01-29
[SP-4054] Backport of PRD-5932 - When Using 'Row' Layout in PRD, Repo…

pentaho-lmartins:
…rt Elements Set to 'visible=false' Still Display on Excel or Excel 2007 Output (8.0 Suite)

@pamval @duarteteixeira


pamval:
why didn't you backport the unit test?

pentaho-lmartins:
@pamval Added test cases from master to this branch.

PR #1093, opened 2018-01-31
[PRD-4595]-Errors from the datasource used in a SINGLEVALUEQUERY function are consumed silently

stepanovdg:
Fix for the missed function URLBUILDER which is used in drilldown.properties
Some default samples uses drilldown.

Fix for serialization into prpt failonerror property (before it was ignored).

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/modules/parser/base/common/ExpressionReadHandler.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/modules/parser/base/common/ExpressionReadHandler.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2001 - 2013 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
+ * Copyright (c) 2001 - 2018 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
  */
 
 package org.pentaho.reporting.engine.classic.core.modules.parser.base.common;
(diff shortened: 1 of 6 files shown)


stepanovdg:
@tmorgner Could you please take a look?

stepanovdg:
@tmorgner @pamval @mbatchelor Could you please take a look?

tmorgner:
Your change contains a new URLBuilderFunction formula function? That is probably part of a different case and should be moved into its own pull request. It most definitely does not deal with formula error flags, after all. 

Also: There should be a unit test for the saving fix. Validate that the value is saved correctly by creating a master-report instance with formulas, then save the report and then load the report and the actually check the formula you saved.

Have a look at GoldSaveLoadIt#postProcess for an example on how to save and load a report in a unit test.

stepanovdg:
@tmorgner  Thank you for review. I will proceed with unit tests. 

According to URLBuilderFunction. Original issue PRD-4595 wasn`t closed - as after fix was found that some samples stop working. As was changed default behavior for fail on error on server and now errors are not ignored - first approach was to simply disable fail on error for the specific function of the sample. But during the investigation - was found that such parameter is ignored at all and not serializing into prprt. So that`s why was added saving/reading of failOnerror boolean. 
But still after all changes was found that error wasn`t related to user defined function - and we were not able to ignore errors for that function. 
For some reason in drill-down config file for the scope of the local-prpt was added URLBuilderFunction call (by David Kincade in 2013). I hope that at that moment such function had to be available. But now it disappeared. I wasn`t able to find commit or any other history about such function - so that why decide to  rewrite it. If anybody has more information about https://github.com/pentaho/pentaho-platform/blob/9f2d7007963d452096ddb81b7a4190ee8f181fda/assemblies/pentaho-war/src/main/webapp/WEB-INF/classes/drilldown-profile.xml#L32 that function please share it.

tmorgner:
Wow, you are right. Those functions do not seem to exist in the BI-server anywhere (and I cannot seem to find references to them in the older code). I retract my objections. Merge ahead!

PR #1096, opened 2018-02-06
[PRD-2943] Hitting save on subreport tab does not reset the change flag for the other subreports and master reports.

Advent51:
- only master report own a valuable changeTracker field - all subreports use the master report's one.
- tests were added


Advent51:
@tmorgner please review

tmorgner:
Cool, you created a wormhole. Coverage went up (according to the yellow icon) and down at the same time. Apart from messing up the time-space continuum, this is good code and merged.

PR #1102, opened 2018-02-26
[SP-3918][PRD-4595]-Errors from the datasource used in a SINGLEVALUEQUERY function are consumed silently Fix for the missed function URLBUILDER which is used in drilldown.properties Some default samples uses drilldown.

stepanovdg:


stepanovdg:
@mchen-len-son Could you please take a look?

PR #1107, opened 2018-03-05
[PRD-5952] PRPT : Row Limit "No More Than" is not getting applied to Sub reports.

Advent51:
- queryLimit will be used form the parent state if it is provided and from subreport's attributes otherwise

### DIFF
--- a/designer/report-designer/src/main/java/org/pentaho/reporting/designer/core/actions/global/NewReportAction.java
+++ b/designer/report-designer/src/main/java/org/pentaho/reporting/designer/core/actions/global/NewReportAction.java
@@ -12,7 +12,7 @@
 * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Lesser General Public License for more details.
 *
-* Copyright (c) 2002-2017 Hitachi Vantara..  All rights reserved.
+* Copyright (c) 2002-2018 Hitachi Vantara.  All rights reserved.
 */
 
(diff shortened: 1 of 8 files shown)


Advent51:
@tmorgner @pamval please review

tmorgner:
Also: This is a very subtle change and it can alter reports in a rather unpredictable way (think of a master report of N customers but you want to show all their orders). 

Please wrap that into a feature toggle so that old reports dont break silently by showing incomplete data. I feel our bank and other regulated customer would not like the outcome of that. The default should be the safe old behaviour.

As for the feature toggle: Make it a new boolean attribute on the master-report and sub-report. If the attribute is null, look for the value in the parent-report, or false when this is the master-report. If set to false, use the old behaviour of not limiting the reports. If the attribute is true, limit the subreports as you plan to do here. 

Now, to make sure all *new* reports get that toggle, you can add the necessary code into PRD's NewReportAction#prepareMasterReport(). This way old reports will behave correctly even if they are edited in PRD, but any new report will have the new behaviour active. Any user who wants to have his existing reports join in with the new behaviour can do so by setting the attribute on the master report. 

And by having the feature-flag inherited from the parent report, you can even limit it for most reports, but opt-out from the limits on certain subreports.

Advent51:

tmorgner:
Sweet and to the point. 

Not sure what you said to Wingman to make this robot complain, but I love the fix and merge it anyway. Wingman apparently took its sensitivity training way to serious and now is offended by unrelated stuff (like coverage changes in code you did not touch). Machine overlords! Bah!

PR #1110, opened 2018-03-08
[PRD-5956] - PRD parameter viewer not handling timezones correctly

CravoBranco:
* Always set the timezone and disable the lenient parsing flag when a date format is created.


CravoBranco:
@pamval , @mchen-len-son , @mbatchelor , @pentaho-lmartins , @tmorgner please review and merge if it's all ok. Thanks in advance!

tmorgner:
Looks good to me.

PR #1113, opened 2018-03-20
[PRD-4692] Fix FormulaEditorPanelTests - One fails always, one randomly

ricardosilva88:
@tmorgner, as request on https://jira.pentaho.com/browse/PRD-4692


tmorgner:
Thank you!

PR #1116, opened 2018-03-22
[SP-4171] Backport of PRD-2943 - Hitting save on subreport tab does not reset the change flag for the other subreports and master reports. (7.1 Suite)

Advent51:
- backport of #1096


Advent51:

PR #1124, opened 2018-04-03
[PRD-5435]-Excel Output - Dynamic height setting is not applied

PashkovichAndrei:


tmorgner:
Please add a config description (in ../modules/output/table/xls/config-descriptions.xml ) for the new property. (Make it easy on yourself, copy one of the existing boolean key declarations and simply update the name and description.) Set the "global" flag to true to indicate that this is a system wide setting.

This is an easily validatable thing, so it might be a good idea to add an integration level test here too. Create a new MasterReport, add a label at (x:0, y:0) into the report-header with some text and a known font, the export to excel and use POI to read the first cell and check its height. Use "Arial" as that font is easy to validate by humans and is also contained in the recorded font metrics in the integration tests (means the CI server sees the correct values regardless of the fonts installed locally).

ricardosilva88:
I will carry on this PR since Andrei Pashkovich left Pentaho

mbatchelor:
Thomas' review on PRD-5435 shows that this fix was insufficient. This is closed in favor of #1129

PR #1125, opened 2018-04-04
[PRD-5988] In reopened Hyperlink Dialog the 'Location' selector on the top left is set to 'Self' instead of the originally selected 'Pentaho Repository'

ricardosilva88:
This commit will fix both PRD-5988, PRD-5970 and PRD-5971. It was very hard to keep PRD working if both changes had been made separately, so I did both changes in same commit. 

Major changes are: Corrected drilldown default profile configuration; Added an handler to set ::pentaho-path on drill down formula and a Parameter filter mechanism to avoid empty ("NA()") system params on Drill down formula (which causes problems on opening hyperlinks to other reports). Some other minor adjustments were made on swing drill down classes. 

@tmorgner


tmorgner:
The only thing wrong I can spot is a distinct lack of unicorn dust. Otherwise: Totally mergable!

PR #1128, opened 2018-04-06
[BACKLOG-22676] Errors displays when publishing report from PRD to server

ricardosilva88:
@tmorgner 

To fix the problem introduced by PRD-5569

### DIFF
--- a/designer/report-designer-extension-pentaho/src/main/java/org/pentaho/reporting/designer/extensions/pentaho/repository/actions/PublishToServerTask.java
+++ b/designer/report-designer-extension-pentaho/src/main/java/org/pentaho/reporting/designer/extensions/pentaho/repository/actions/PublishToServerTask.java
@@ -82,7 +82,10 @@ public class PublishToServerTask implements AuthenticatedServerTask {
 
       //populate all properties from file which loaded in report designer before publish it to server
       Properties fileProperties = new Properties();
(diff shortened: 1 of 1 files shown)


designer/report-designer-extension-pentaho/src/main/java/org/pentaho/reporting/designer/extensions/pentaho/repository/actions/PublishToServerTask.java
```diff
@@ -82,7 +82,12 @@ public void run() {
 
       //populate all properties from file which loaded in report designer before publish it to server
       Properties fileProperties = new Properties();
-      fileProperties.setProperty( PublishRestUtil.REPORT_TITLE_KEY, selectFileForPublishTask.getReportTitle() );
+      String reportTitle = selectFileForPublishTask.getReportTitle();
+      if ( reportTitle != null ) {
+        fileProperties.setProperty( PublishRestUtil.REPORT_TITLE_KEY, reportTitle );
+      } else {
+        fileProperties.setProperty( PublishRestUtil.REPORT_TITLE_KEY, "" );
```
tmorgner: Not sure whether the BI-server would handle an empty title gracefully. A non existing title, yes, but empty string. We might end up with an report that displays itself as empty string. Might be better to simply do nothing in the else branch, and thus not set the property at all.

ricardosilva88:
@tmorgner 

Updated PR. Tested and looks good.

PR #1129, opened 2018-04-11
[PRD-5435] Excel Output - Dynamic height setting is not applied

ricardosilva88:
@tmorgner

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/text/TextUtility.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/text/TextUtility.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2001 - 2013 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
+ * Copyright (c) 2001 - 2018 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
  */
 
 package org.pentaho.reporting.engine.classic.core.layout.text;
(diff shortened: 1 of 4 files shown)


engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/text/TextUtility.java
```diff
@@ -62,4 +62,27 @@ public static ExtendedBaselineInfo createBaselineInfo( final int codepoint, fina
         fontMetrics.getMaxHeight(), underlinePosition, strikeThroughPosition );
   }
 
+  public static ExtendedBaselineInfo createCorrectBaselineInfo( final int codepoint, final FontMetrics fontMetrics,
```
tmorgner: I'd name this function 'createPaddedBaselineInfo' instead. If you say create correct bla, then I'd wonder is the the other method not correct? 

Also: This method is unlogical for the causal reader (head voice: "why the heck would they do that?!") so it might be worthwhile to add a extra comment here:

"This method is a workaround for broken font metrics supplied by some truetype fonts. Those fonts trip up Excel's row heights in the workbooks. We therefore try to detect those fonts by checking whether their declared sizes are awfully close to the requested font size and apply some extra padding in that case. For all details please read PRD-5435."

This just makes sure that the next person touching that code gets an idea of the madness that goes on without having to dig into a long conversation chain on the case. This then also gives context to the magic numbers you have in your code (and thus frees you from having them as named constants, which for a single use like that always feels over the top.

engine/core/src/test/java/org/pentaho/reporting/engine/classic/core/layout/text/TextUtilityTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.pentaho.reporting.libraries.fonts.registry.FontMetricsFactory;
+import org.pentaho.reporting.libraries.fonts.registry.DefaultFontContext;
+import org.pentaho.reporting.libraries.fonts.registry.FontMetrics;
+import org.pentaho.reporting.libraries.fonts.registry.FontFamily;
+import org.pentaho.reporting.libraries.fonts.registry.FontRecord;
+
+
+public class TextUtilityTest {
+
+  @Test
```
tmorgner: I would not put that test in here like that. You are testing against the runtime environment and you really cannot make any assumptions about the fonts they may have installed. I only added that 'non-test' in my example to demonstrate the fonts that may be bogus and to allow you to replicate these conclusions on your system. 

Your test will fail on a perfect system that has no bad fonts.

However, what you can test the function more reasonably by passing a mock of a fontmetrics object in and validate the outcome both both cases (one for a good font, one for a bad font).

As the Excel output target does not allow us to inject our own debug-level font contex into the process, there is no safe way to isolate the test from the runtime (so that we dont rely on fonts produced by the native AWT) so there is no safe way to create integration tests that aren't bogus or prone to failure. So the excel export for this case must be validated manually by QA later.

ricardosilva88:
@tmorgner  PR updated with your feedback

engine/core/pom.xml
```diff
@@ -241,6 +241,10 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
     </dependency>
+    <dependency>
```
tmorgner: No need for that here ...

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/text/TextUtility.java
```diff
@@ -29,6 +30,7 @@
   private TextUtility() {
   }
 
+  @VisibleForTesting
```
tmorgner: .. nor that ...

engine/core/src/test/java/org/pentaho/reporting/engine/classic/core/layout/text/TextUtilityTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+import org.junit.runner.RunWith;
+
+
+@RunWith( PowerMockRunner.class )
+@PrepareForTest( TextUtility.class )
+public class TextUtilityTest {
+
+  @Test
```
tmorgner: .. because the TextUtilityMethod is a pure function and only depends on the inputs that you pass in. 

The function you want to test is the 'createPaddedBaselineInfo' which takes two parameters that are complex. Ignore the 'BaselineInfo' parameter, as this is a final class and just a data carrier. Just instantiate one and set some sensible values there.

The only thing you need to actually mock is the font metrics object. "FontMetrics#getBaselines" can simply return your BaselineInfo object from above, and the other three methods used can be easily stubbed, as you do in the code here. 

The static method that you annotated with that Guava annotation does not need to annotated at all. Its behaviour is controlled by the dominant-baseline method returned via the BaselineInfo object. Just set that property to 'MIDDLE' in your BaselineInfo object (that you return as pass-through value from the font metrics mock) when you set up the object and it will return the right value for you. In any case, that dominant baseline is only passed on to the resulting ExtendedBaselineInfo and has no bearing on the logical flow of your method under test. So set up your tests that it returns something sensible, but the exact details of that value wont matter.

tmorgner:
Looks good.

PR #1132, opened 2018-04-16
[PRD-5989]-Horizontal line is not visible in HTML when the line is dashed, dotted, dot-dash or dot-dot-dashed

denisprotopopov:
-map to the nearest equivalent


denisprotopopov:

PR #1126, opened 2018-04-05
[PDB-2004] SQL List Prompts are converting parameters based on string to double precision numbers

Advent51:
- new conditions for the String parameters processing were added
- new tests were added

PR #1136, opened 2018-04-24
[SP-4284] Backport of PDB-2004 - SQL List Prompts are converting parameters based on string to double precision numbers (8.0 Suite)

Advent51:
- backport of #1126


Advent51:
@tmorgner please review

PR #1159, opened 2018-06-14
[BACKLOG-23781] PRD is not returning DB connection detail and its not…

ricardosilva88:
… possible to execute querys

@tmorgner 
@mbatchelor 
@lucboudreau 

Lost's of code formatted due to checkstyle problems.

Code changes are on:
JdbcDataSourceDialog - Line 136 to 143 - Getting report parameters
JdbcPreviewWorker - Most code changes - Receiving parameters, treat them and pass to query treatment flow

The other java class (JdbcQueryDesignDialog l.55) is a side effect of function signature change.

### DIFF
--- a/designer/datasource-editor-jdbc/src/main/java/org/pentaho/reporting/ui/datasources/jdbc/ui/JdbcPreviewWorker.java
+++ b/designer/datasource-editor-jdbc/src/main/java/org/pentaho/reporting/ui/datasources/jdbc/ui/JdbcPreviewWorker.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2009 - 2017 Hitachi Vantara.  All rights reserved.
+ * Copyright (c) 2009 - 2018 Hitachi Vantara.  All rights reserved.
  */
 
 package org.pentaho.reporting.ui.datasources.jdbc.ui;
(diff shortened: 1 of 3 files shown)


designer/datasource-editor-jdbc/src/main/java/org/pentaho/reporting/ui/datasources/jdbc/ui/JdbcPreviewWorker.java
```diff
@@ -88,19 +108,19 @@ public void cancelProcessing(final CancelEvent event)
...
+      if ( this.reportParameters != null ) {
+        paramDataRow = normalizeParameters( this.reportParameters );
       }
-    }
-    catch (ReportDataFactoryException e)
-    {
+
+      tableModel = dataFactory.queryData( query, new QueryDataRowWrapper( paramDataRow, queryLimit, queryTimeout ) );
+
+      if ( queryLimit > 0 ) {
```
lucboudreau: This should be greater or equal to 0. A 0 limit must also be enforced.

tmorgner: Nope. We are using JDBC semantics, therefore zero means no limit. See https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html#setMaxRows(int)

ricardosilva88: I hadn't change anything there @lucboudreau. The code changes that I did were:
JdbcDataSourceDialog - Line 136 to 143 - Getting report parameters
JdbcPreviewWorker - Most code changes - Receiving parameters, treat them and pass to query treatment flow
The other java class (JdbcQueryDesignDialog l.55) is a side effect of function signature change.

PR #1160, opened 2018-06-18
[PRD-4896] SQL Query Designer Preview button not working

ricardosilva88:
@tmorgner 

**Note for future reference (SP's process): This commit uses code made in context of https://jira.pentaho.com/browse/BACKLOG-23781 (already merged), please respect this order**

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/modules/misc/datafactory/sql/SimpleSQLReportDataFactory.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/modules/misc/datafactory/sql/SimpleSQLReportDataFactory.java
@@ -62,10 +62,15 @@ public class SimpleSQLReportDataFactory extends AbstractDataFactory {
   private String userField;
   private String passwordField;
 
+  public static Configuration globalConfig;
+
   public SimpleSQLReportDataFactory() {
-    final Configuration globalConfig = ClassicEngineBoot.getInstance().getGlobalConfig();
-    this.columnNameMapping = "Name".equalsIgnoreCase( globalConfig.getConfigProperty( //$NON-NLS-1$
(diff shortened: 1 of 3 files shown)


engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/modules/misc/datafactory/sql/SimpleSQLReportDataFactory.java
```diff
@@ -130,9 +136,10 @@ protected synchronized Connection getConnection( final DataRow dataRow ) throws
     return connection;
   }
 
-  private int getBestResultSetType( final DataRow dataRow ) throws SQLException {
-    if ( "simple".equalsIgnoreCase( getConfiguration().getConfigProperty( //$NON-NLS-1$
-        ResultSetTableModelFactory.RESULTSET_FACTORY_MODE ) ) ) { //$NON-NLS-1$
+  @VisibleForTesting
```
tmorgner: Is this Guava thing really necessary? This is a protected method that is only referenced from the same package (as the tests are local), so this should not be needed. Putting a new dependency in just for the sake of test annotations seems so wrong on so many levels. 

To Be honest, that method could even be public static, as the "globalConfig" variable you need from the local instance is actually fetched from a static context anyway. If you inline this, that becomes clear. And for static helper methods public does not break encapsulation so that is fine by all standards I'm aware of.

ricardosilva88:
@tmorgner , just updated the PR. Removed the "Visible for testing" annotation, changed the getBestResultSetType method to public and globalConfig var to public static. Didn't put static on getBestResultSetType method because there are other methods that would be need a cascade update of static modifier, like getConnection.

ricardosilva88:
@tmorgner  this PR depends on code made in context of https://jira.pentaho.com/browse/BACKLOG-23781, that wasn't fully merged, it is missing one PR - https://github.com/pentaho/pentaho-kettle/pull/5487

tmorgner:
Looks sane!

PR #1169, opened 2018-07-26
[PIR-1172] - L10N - Untranslated string in PIR prompt.

cravobranco:
* Messages handling mechanism changed from singleton to multi instance.
* Pentaho's platform configured locale is now used to get the correct language.

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/parameters/Messages.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/parameters/Messages.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2017 Hitachi Vantara..  All rights reserved.
+ * Copyright (c) 2002-2018 Hitachi Vantara..  All rights reserved.
  */
 
 package org.pentaho.reporting.engine.classic.core.parameters;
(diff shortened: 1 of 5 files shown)


cravobranco:
@pentaho-lmartins , @tmorgner , @pamval , @graimundo , @mbatchelor please review and merge if it's all ok.

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/parameters/Messages.java
```diff
@@ -12,33 +12,38 @@
...
 
 public class Messages extends ResourceBundleSupport {
-  private static Messages instance;
 
-  public static Messages getInstance() {
-    // its ok that this one is not synchronized. I dont care whether we have multiple instances of this
-    // beast sitting around, as this is a singleton for convinience reasons.
-    if ( instance == null ) {
-      instance = new Messages();
+  private static Map<Locale, Messages> messagesMap = new HashMap<>();
```
tmorgner: This caching is not necessary. The locale handling is lightweight, as the actual IO is cached deep inside ResourceBundle itself (all handled by the JDK; and aside from that, the methods here are never called from a performance critical context anyway).

However, if  you'd want to do caching, you'd have to do that in a more sane way where you take care of race conditions when using mutable structures on a static context. Without synchronisation your hashmap wont like it if it is called from two different threads at the wrong time (at some point that would happen in the server, and debugging that would be fun). But if you synchronize, you introduce a global lock (static synchronised methods are a nightmare!) which might now slow down the server more than what you'd pay if you were to simply create multiple objects.

So if in doubt, use the standard rule: dont optimize until you know you have a problem in that particular code. 

And for this instance, if you call these getInstance() methods so many times that it becomes a problem, it means that somebody is hammering the server with parameter-validation requests. And dealing with those will be by far more expensive than your translation message handling ever could be (as database access and XML parsing trumps pretty much anything else).

cravobranco: @tmorgner , thanks for your feedback!
@graimundo , could you please comment the remark above?

From my side it's ok to revert this part to the previous implementation, however I would like to know if you have any objection to drop this "singleton -> multi instance" approach.

cravobranco: @tmorgner  , The new structure introduced in Messages.java is not related to caching but to multi-instance (vs singleton) approach.

If we keep Messages.java as a singleton, the following problem will happen (I verified it by myself):

1) After the first login, a user configures his session language as French.
2) The user then opens an Interactive Report and sees the message as French, as far singleton instance of Messages.java is instantiated with locale=French, OK as expected.
3) Then the same user configures his session language as German, opens an IR and sees that the messages are still in French, NOT OK.

The reason for the wrong behavior is the singleton approach of Messages.java which keeps the first configured locale forever and ever because Messages instantiation is unique.
The committed code on this PR fixes it.

tmorgner: Dont keep it as singleton, let "getInstance()" return a new instance on every call. There is no need for a Hashmap holding instances (which is a cache in all but name!) The Messages objects are very lightweight so there is no need to avoid having multiple instances. 

Thus a 

       getInstance(Locale l) {
         return new Messages(l); 
       }

will do the trick without any extra code for holding instances longer than needed.

cravobranco:
@tmorgner , thanks for your feedback!
As agreed, I followed your suggestion and implemented your proposal.

No problems were found when I tested the edge case scenario.
Therefore, please review again this PR and merge if it's all ok.

PR #1171, opened 2018-07-27
[PRD-4907] When using text/html on a text-field no more style can be …

ricardosilva88:
…applied on the object

@tmorgner

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/build/RichTextStyleResolver.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/build/RichTextStyleResolver.java
@@ -12,7 +12,7 @@
  *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  *  See the GNU Lesser General Public License for more details.
  *
- *  Copyright (c) 2006 - 2017 Hitachi Vantara..  All rights reserved.
+ *  Copyright (c) 2006 - 2018 Hitachi Vantara..  All rights reserved.
  */
 
 package org.pentaho.reporting.engine.classic.core.layout.build;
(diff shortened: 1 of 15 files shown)


pamval:
@tmorgner can you review this pr, please?

ricardosilva88:
@tmorgner , as discussed via Slack, i just updated the PR with your feedback

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/build/RichTextStyleResolver.java
```diff
@@ -21,35 +21,40 @@
...
+import org.pentaho.reporting.engine.classic.core.style.css.CSSStyleResolver;
+import org.pentaho.reporting.engine.classic.core.style.resolver.StyleResolver;
 import org.pentaho.reporting.engine.classic.core.util.AbstractStructureVisitor;
+import org.pentaho.reporting.engine.classic.core.function.ProcessingContext;
 
 public class RichTextStyleResolver extends AbstractStructureVisitor {
-  private SimpleStyleResolver simpleStyleResolver;
-  private ResolverStyleSheet resolveStyleSheet;
+  private StyleResolver cssStyleResolver;
+  private static RichTextStyleResolver richResolverInstance;
```
tmorgner: This variable is not used anymore, isnt it? However, given that the resolver instance is now tied to the report-process, and given that the report-process is guaranteed to be single-threaded, you can bring back the cached instance of the resolveStyleSheet instead of creating a new one every time. This thing is an aggregator that then gets copied into the SimpleStyleSheet when resolving is complete. Reusing the same instance of it means you dont have to create a new instance (and all its internal arrays) for each element you validate.

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/modules/output/pageable/graphics/internal/LogicalPageDrawable.java
```diff
@@ -766,7 +766,7 @@ protected void finishInlineBox( final InlineRenderBox box ) {
         strikeThrough = null;
       }
     } else {
-      underline = computeUnderline( box, null );
+      strikeThrough = computeStrikeThrough( box, null );
```
tmorgner: Nice catch!

ricardosilva88: ty :)

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/build/RichTextStyleResolver.java
ricardosilva88: Yeap, I totally forgot to remove it. I'll update the PR.

ricardosilva88:
@tmorgner PR updated

PR #1173, opened 2018-08-13
[PRD-6010] Unable to perform simple calculations using count, sum and…

LeonardoCoelho71950:
… openformula

If an open-formula contains a count or a sum function, the underlying functions must run before the open-formula, so that the open-formula has the proper values for calculation purposes. 

@pamval @duarteteixeira 
@pentaho-lmartins @ricardosilva88

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/ItemPercentageFunction.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/function/ItemPercentageFunction.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2001 - 2013 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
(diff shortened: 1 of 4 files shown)


pamval:
@tmorgner can you please take a look at this pull request?

tmorgner:
Hardcoding the dependency level to 1 is not a valid solution. Its may be a workaround, but not something I'd want to see in production. Doing so blindly will break other reports where users actually declared a valid set of dependencies between their functions. 

The problem here is that the two functions are "Total*Functions", which means they do not have a valid result until the dependency level calculations have finished. Until then they probably return zero. Which, as the formula does not actually check the value of countTest, results in a DIV/0 error. 

In the old version, we did not report errors, so that error was silently ignored (but it occurred nonetheless). Now since  PRD-4595 those errors are reported back. 

There is nothing to fix, this error is normal for the situation. To ignore this error on this function, use the "failOnError" property on the formula function itself. 

If the customer want to revert to the old global behaviour of ignoring errors, let them set the property "org.pentaho.reporting.engine.classic.core.StrictErrorHandling" to false, either globally via the classic-engine.properties file or on a per report basis via the report-configuration.

pentaho-lmartins:
Closing and adding the notes above to the JIRA case.

PR #1185, opened 2018-09-14
[SP-4589] - Backport of PRD-5628 - Report Wizard: The check box Disti…

cravobranco:
…nct Only is hidden on Format step (7.1 Suite)

* Backport of PRD-5628 - Report Wizard: The check box Distinct Only is hidden on Format step (7.1 Suite)

@pentaho-lmartins , @ricardosilva88


cravobranco:
This PR is no more valid because base case (PRD-5628) was reopened.

PR #1187, opened 2018-10-08
[PDI-3493] Variables not replaced in MDX from Mondrian input step whe…

ppatricio:
…n called from Reporting Engine

We have to set the parameter value also into the TransMeta object.
@pentaho-lmartins


tmorgner:
Looks good. Merged.

PR #1190, opened 2018-10-09
[PRD-5009] Formula editor does not handle complex terms properly

ppatricio:
The open parenthesis characters were being normalized as FormulaFunctionElement's and not as FormulaOpenParenthesisElement's - bugging all the open and close parenthesis logic. 
This PR adds that clause, ensuring that.
@tmorgner

### DIFF
--- a/libraries/libformula-ui/src/test/java/org/pentaho/openformula/ui/FormulaDocumentTest.java
+++ b/libraries/libformula-ui/src/test/java/org/pentaho/openformula/ui/FormulaDocumentTest.java
@@ -12,16 +12,25 @@
 * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Lesser General Public License for more details.
 *
-* Copyright (c) 2002-2017 Hitachi Vantara..  All rights reserved.
+* Copyright (c) 2002-2018 Hitachi Vantara..  All rights reserved.
 */
 
 package org.pentaho.openformula.ui;
 
 import junit.framework.TestCase;
+import org.pentaho.openformula.ui.model2.FormulaClosingParenthesisElement;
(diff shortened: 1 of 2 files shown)


tmorgner:
Looks very good. 

Can you add a quick unit test that validates this parsing? A quick "FormulaParser#parse" with a few asserts to look for the parenthesis token would make sure this stuff never ever breaks again.

ppatricio:
Hey Marc,
only saw your comment after updating the PR. Is something like this you want? 
But with more covered syntax?

libraries/libformula-ui/src/test/java/org/pentaho/openformula/ui/FormulaDocumentTest.java
```diff
@@ -57,4 +62,23 @@ public void testParseFunctionBad2() throws BadLocationException {
     assertEquals( "Length", str.length(), doc.getLength() );
     assertEquals( "Number of elements: ", 17, element.getElementCount() );
   }
+
+  public void testParseTextOpenParenthesisCorrectlyPRD5009()  {
+    FormulaDocument doc = new FormulaDocument();
+    final String str = "=IF((1=3);1;2)";
+    FormulaElement[] result = FormulaParser.parseText( doc, str );
+    assertEquals( "class of =", result[0].getClass().getName(), "org.pentaho.openformula.ui.model2.FormulaOperatorElement" );
```
tmorgner: Instead of using text here, compare the class instances directly. Class instances are equal when loaded from the same classloader, something that is guaranteed to be true for JUnit runners. So kill "getName()" and use FormulaOperatorElement.class and so on ..

ppatricio: Thanks Thomas, PR Updated. 
Tried not to import all those classes, but it definitely looks better this way.

PR #1193, opened 2018-10-15
[PPP-4167] - Assembly “pre-classic-sdk" seems to be missing since 7.0

ricardosilva88:
@jmcrfp need your review on this one and please add sdk module to building page - pre-classic-sdk-8.2.0.0-SNAPSHOT.zip

@tmorgner your feedback is also important on this one

I'll also open a ticket to update the pdf document of this project - https://github.com/pentaho/pentaho-reporting/blob/6.1/engine/sdk/resources/documentation/embedding_pentaho_reporting_engine.pdf


sql/sampledata.log
```diff
@@ -0,0 +1,6 @@
+/*C4*/SET SCHEMA PUBLIC
```
tmorgner: Please gitignore this "*.log" file or any developer that runs the build or samples will modify it needlessly. HSQL will regenerate it if it is missing without any problems.

sql/sampledata.properties
```diff
@@ -0,0 +1,4 @@
+#HSQL Database Engine 2.3.2
```
tmorgner: Dito. Kill that one. It will be regenerated.

ricardosilva88:
@tmorgner , just updated PR... didn't noticed that those db files where there. They are already on  \assemblies\sdk\src\main\resources\samples\sql\

jmcrfp:
@pentaho/x-wing

tmorgner:
Looks good to me.

PR #1204, opened 2018-10-31
[BACKLOG-26393] Report Designer hyperlinking not working

pedrofvteixeira:
1.  With the work done for BISERVER-13648, the server no longer return an auth challenge to the issued request ( when this used to be the case, the HttpClient lib would handle such a 401 response OOTB by issuing the credentials should any have been set in the HTTPContext ).
2. Now ( i.e. post BISERVER-13648 ), the server returns a 302 REDIRECT to the server's /Login page.
3. We can circumvent by defining an  HttpClientContext with preemptive authentication set.
4. Note that we're only setting preemptive auth for this request and no others, given that this is the only request that goes through the server's `/api/repos/**` spring-security regex. Other requests are working as intended, but are issued towards other regexs, for which BISERVER-13648 did not change their behaviour

@pamval @pentaho-lmartins please process

**For clarity**

This has been tested and successfully validated using both an `applicationContext-spring-security.xml` with and without the changes made for BISERVER-13648. 
  - Considering a scenario where, through an upgrade path, an older version of that config file gets ported over ( i.e. one that still declares the old way of using `exceptionContextFilterForWS` on `/api/repos/**` regex.

### DIFF
--- a/designer/report-designer-extension-pentaho/src/main/java/org/pentaho/reporting/designer/extensions/pentaho/drilldown/PentahoParameterRefreshHandler.java
+++ b/designer/report-designer-extension-pentaho/src/main/java/org/pentaho/reporting/designer/extensions/pentaho/drilldown/PentahoParameterRefreshHandler.java
@@ -290,8 +290,8 @@ public class PentahoParameterRefreshHandler implements DrillDownParameterRefresh
      */
     private HttpClientContext buildPreemptiveAuthRequestContext( final URI target, final AuthenticationData auth ) {
 
-      if ( target == null || auth == null ) {
-        return null; // nothing to do here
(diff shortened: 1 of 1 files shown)


pedrofvteixeira:
@pamval @pentaho-lmartins please review / process. I'll do the the unit tests at a later time ( no sense in holding off getting this merged in and into a nightly master QAT build )

pentaho-lmartins:
Another approach would be to implement a custom filter extending from the spring framework filter that would either redirect to the login page or send an authentication challenge to the client based on the user agent, provided that PRD and all other client tools would have a specific user agent that would identify them as _our_ client tools, and as such don't behave as a web browser.

This would be a more elegant solution, but given the time frame it may not be feasible.

PR #1213, opened 2018-12-04
[PPP-4167] - Linux fix on assembly "pre-classic-sdk"

ricardosilva88:
Fix's to bugs found by Vitor when trying to back port base correction to 7.1

@cardosov

### DIFF
--- a/assemblies/samples/src/assembly/assembly.xml
+++ b/assemblies/samples/src/assembly/assembly.xml
@@ -22,7 +22,16 @@
     <fileSet>
       <directory>.</directory>
       <outputDirectory>.</outputDirectory>
+      <includes>
+        <include>*.sh</include>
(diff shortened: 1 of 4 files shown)


assemblies/samples/src/assembly/assembly.xml
```diff
@@ -26,6 +26,7 @@
         <exclude>/src/</exclude>
         <exclude>/target/</exclude>
       </excludes>
+      <fileMode>0755</fileMode>
```
cardosov: This is going to give execution permissions to all files, not just the shell scripts. Is this intentional? Comparing to the files from the 7.0 artifact, only the '.sh' files have the execution permissions.

ricardosilva88:
@cardosov PR updated

PR #1233, opened 2019-02-18
[PIR-1421] Unexpected space appears when aligning columns in Interact…

LeonardoCoelho71950:
…ive report

@pentaho/tatooine @pentaho-lmartins 
@tmorgner 

Hi @tmorgner, I made a new change in DefaultTextExtractor, because if I add a zero width space between words they would be coupled together. For instance:
  "Orde\u200Br\u200BNum\u200Bber" => OrderNumber (Incorrect)
  "Orde\u200Br Num\u200Bber" => Order Number (Correct)


engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/modules/output/table/base/DefaultTextExtractor.java
```diff
@@ -390,7 +390,17 @@ protected void processParagraphChilds( final ParagraphRenderBox box ) {
         if ( lineBox.getStaticBoxLayoutProperties().isPreserveSpace() == false ) {
           addSoftBreak();
         } else {
-          addEmptyBreak();
+
+          RenderableText renderableText = (RenderableText) lineBox.getFirstChild();
```
tmorgner: You cannot assume that the first child will be a renderable text element. It could also be a renderbox, or an image for instance. You have to be more defensive here. In case there are more than one child elements, ask yourself: Do you want to test the first child of the line to determine the type of space or the last child? 

I'd say recursively search the linebox for the deepest last child and base your decision on that. So if last child is box, search recursively within the box, if last child is not text, assume soft break, else use your logic below.

LeonardoCoelho71950: @tmorgner I've updated the PR. Now it recursively searches for the deepest not box last child.
If that element is a RenderableText, it decides between soft or empty break.
If not, it just put a soft break.

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/modules/output/table/base/DefaultTextExtractor.java
```diff
@@ -390,13 +390,35 @@ protected void processParagraphChilds( final ParagraphRenderBox box ) {
...
+          }
         }
       }
       lineBox = (RenderBox) lineBox.getNext();
     }
   }
 
+  protected RenderNode getDeepestLastChild( RenderBox lineBox ) {
+    if ( lineBox.getLastChild() != null && lineBox.getLastChild() instanceof RenderBox ) {
+      return getDeepestLastChild( lineBox );
```
tmorgner: Infinite loop? You probably wanted to pass a cast version of lineBox.getlastChild instead.

LeonardoCoelho71950: Ups, that's right!
PR updated.

tmorgner:
Merged.

PR #1237, opened 2019-02-25
[BACKLOG-27567] Remove org.osgi.core version override

nantunes:
Merge after https://github.com/pentaho/maven-parent-poms/pull/105.
@graimundo please review.

### DIFF
--- a/assemblies/prd-ce/pom.xml
+++ b/assemblies/prd-ce/pom.xml
@@ -176,19 +176,10 @@
       <scope>compile</scope>
     </dependency>
 
-    <!-- TODO: When replacing felix framework implementation for OSGi core specification 4.3.1 which is the spec
-         we aim at (see pentaho-parent-pom ) in our Karaf 3.0.3 assembly an error occurs:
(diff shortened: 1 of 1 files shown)


assemblies/prd-ce/pom.xml
```diff
@@ -175,22 +175,6 @@
...
-    <!-- TODO: When replacing felix framework implementation for OSGi core specification 4.3.1 which is the spec
-         we aim at (see pentaho-parent-pom ) in our Karaf 3.0.3 assembly an error occurs:
-         ERROR: Bundle org.ops4j.pax.url.mvn [1] Error starting mvn:org.ops4j.pax.url/pax-url-aether/2.3.0
-         (java.lang.NoSuchMethodError: >org.osgi.framework.wiring.BundleWire.getProvider()
-
-         In this assembly we are using Felix Framework 4.2.1, which is partially OSGi 5.0 complaint. In particular,
-         it is complaint with the org.osgi.framework.wiring.BundleWire.getProvider() method.
-         For this reason we are explicitly overriding OSGi core to 5.0.0 here.
-         -->
-    <dependency>
```
graimundo: @nantunes doesn't the dependency need to be declared to be included?

nantunes: Yes, you're right. I'll fix it.

PR #1238, opened 2019-02-25
[PRD-6049] Report Sub Title 1 is not being exported in Excel [Excel +…

ppatricio:
… Excel (2007)]

This case was a side-effect of https://github.com/pentaho/pentaho-reporting/pull/1129.
Adding extra padding was causing cell overlap on Excel, since the font space reserved (plus extra padding) was bigger than the initial reserved, causing, in this case, the "sub title 1" omission.
Added a new logic to apply the previous fix only when "getMaxHeight() <= getMaxAscent() + getMaxDescent()". 

@tmorgner could you please review this? This actually fixes both cases but I'm not sure if it is enough.

@ricardosilva88 @pentaho-lmartins


tmorgner:
Look totally safe to merge. It would look even more beautiful if the test "TextUtilityTest" (see PR #1129 for that file) would be extended to test this new condition.

ppatricio:
Thanks Thomas. PR Updated.

PR #1240, opened 2019-03-11
[BACKLOG-26682] - As a report designer, i would like to hide parameters based on a formula

hugo-jose:


hugo-jose:
@pentaho/bobafett Please review

PR #1243, opened 2019-03-13
[CLEANUP] Make ReportEnvironmentDataRowTest independent of the system locale

nantunes:
@pentaho/millenniumfalcon please review.


graimundo:
Coverage information makes no sense. The change passes through the same actions.
Merging.

PR #1244, opened 2019-03-25
[BACKLOG-27511] - Karaf upgrade preparation 

graimundo:


graimundo:
Too many interdependencies for wingman. Merging.

PR #1248, opened 2019-03-31
[PDI-15742] Unable to load a prpt from the repository in the Pentaho Reporting Output step

neeta-rathi:
1. Added new implementation of `ResourceData` and `ResourceLoader` for Apache VFS's `FileObject`.

2. Added this new `FileObjectResourceLoader` to `loader.properties`

@tmorgner @pentaho/rogueone 

Note: To be merged with: https://github.com/pentaho/pentaho-kettle/pull/6372

### DIFF
--- a/libraries/libloader/pom.xml
+++ b/libraries/libloader/pom.xml
@@ -149,5 +149,16 @@
       <artifactId>hamcrest-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-vfs2</artifactId>
+      <version>${commons-vfs2.version}</version>
+      <exclusions>
+        <exclusion>
+          <artifactId>*</artifactId>
+          <groupId>*</groupId>
+        </exclusion>
+      </exclusions>
+    </dependency>
   </dependencies>
 </project>
--- a/libraries/libloader/src/main/java/org/pentaho/reporting/libraries/resourceloader/ResourceKey.java
+++ b/libraries/libloader/src/main/java/org/pentaho/reporting/libraries/resourceloader/ResourceKey.java
@@ -12,11 +12,12 @@
 * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Lesser General Public License for more details.
 *
-* Copyright (c) 2006 - 2017 Hitachi Vantara and Contributors.  All rights reserved.
+* Copyright (c) 2006 - 2019 Hitachi Vantara and Contributors.  All rights reserved.
 */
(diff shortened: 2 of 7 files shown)


pedrofvteixeira:
PR looks good.
We should create UTs / ITs to help prove out the fix, as well as to prevent future regressions.

tmorgner:
I sense a distinct lack of testing in this one. Please add a unit test that validates that this actually works and that serialisation works as well. Have a look at the test class "FileResourceLoaderTest" for a template you can use. 

In addition to that it might be good to also test the actual loading process by loading a file. You can simply ask it to load the SVG file that is in the test resources collection, for instance. This ensures that VFS2 plays nicely with the existing code.

neeta-rathi:
@tmorgner I do see a `java.io.NotSerializableException: org.pentaho.platform.repository.solution.filebased.SolutionRepositoryVfsFileObject` logged as an `ERROR` on the logs, however the transformation finishes successfully ( per the steps for repro on jira ), and creates the output.  It looks our implementation of FileObject - `SolutionRepositoryVfsFileObject` needs to be serializable. I'm trying that out, also adding a unit test.

tmorgner:
That's exactly why you need unit tests here and (given that you see the error in Kettle) you probably need to change the new code to make everything properly serializable. I'd assume the internal VFS files are not serializable, so you need to make sure that during serialization you marshall and unmarshall all information needed to restore the object when it is deserialized again.

Without serialization, any report contains elements that references VFS2 links will most likely break. The reporting engine needs serialization to support caching of processed report data - no serialisation means the report processing will be broken.

pedrofvteixeira:
@pentaho/rogueone @tmorgner please review/process updated PR. Thanks

neeta-rathi:
Updated `ResourceKey` to handle resource/identifier of type `FileObject` as it was running into serialization errors for identifier of type `FileObject`. As this PR stands, I see no errors in PS logs and the transformation  runs as expected. This transformation has a Pentaho Reporting Output Step that refers to a prpt on pentaho repo with a path `solution:/home/admin/someName.prpt`

pedrofvteixeira:
@tmorgner @pentaho/rogueone please review/process updated PR. Thanks

neeta-rathi:
@tmorgner `createKey` on `FileObjectResourceLoader` can process an object of type `FileObject` or `String`. All the existing test cases ( `URLResourceLoaderTest`, `ClassloaderResourceLoaderTest`, `XPathQueryTest` ) works as expected. I would be more than happy to update the PR so it works for any edge cases that we have not thought about. Thanks!

tmorgner:
Those tests look good now. Your testing is a heck of a lot more complete than what we had for the older parts of the library; I like it!

PR #1260, opened 2019-04-16
[BACKLOG-26682] As a report designer, i would like to hide parameters based on a formula

hugo-jose:
@pentaho/bobafett please review

### DIFF
--- a/engine/core/pom.xml
+++ b/engine/core/pom.xml
@@ -239,7 +239,6 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>17.0</version>
     </dependency>
   </dependencies>
   <build>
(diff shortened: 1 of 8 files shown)


engine/core/pom.xml
```diff
@@ -236,6 +236,11 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>17.0</version>
```
vflsa: The property ${guava.version} is already defined in the parent pom in pentaho-reporting, you shouldn't explicity put here.

hugo-jose: Done.

PR #1266, opened 2019-05-08
[PRD-6063] Fix libformula ACOSH function to return correct value

cdeptula:
@pentaho/tatooine @pentaho-lmartins @ssamora


smmribeiro:
I'm pushing a commit to add validation on the values that are passed: ACOSH function's domain is [1, +infinity[
Also improving unit testing

PR #1270, opened 2019-05-09
[BISERVER-14206] - Cascade parameter update on report prompt only upd…

ricardosilva88:
…ates once

CC: @pentaho-lmartins

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/cache/CachingDataFactory.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/cache/CachingDataFactory.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2001 - 2013 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
+ * Copyright (c) 2001 - 2019 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
  */
 
 package org.pentaho.reporting.engine.classic.core.cache;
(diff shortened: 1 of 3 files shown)


ricardosilva88:
PR updated to fix tests that were failing

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/cache/CachingDataFactory.java
```diff
@@ -360,7 +360,7 @@ private DataCacheKey createCacheKey( final String query, final DataRow parameter
       DataFactoryMetaData metaData = backend.getMetaData();
 
       final String[] referencedFields = metaData.getReferencedFields( backend, query, parameters );
-      if ( referencedFields != null ) {
+      if ( referencedFields != null && referencedFields.length > 0 ) {
```
tmorgner: This is not correct. With this change you now would no longer cache any unparametrized query at all. Not caching that expensive "SELECT * WHERE x LIKE "%Y%" is going to be bad. Revert this line here.

(And yes, this now means that static table data sources are ending up in the cache, where they may not need to be cached (as they are in memory anyway), but that is better than not caching queries at all. We could address that as a extra setting in the metadata at some future point, but for today, screw it, better be correct than memory efficient.)

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/sorting/SortingDataFactoryCore.java
```diff
@@ -32,7 +32,7 @@ public SortingDataFactoryCore() {
   public String[] getReferencedFields( final DataFactoryMetaData metaData, final DataFactory element,
       final String query, final DataRow parameter ) {
     String[] referencedFields = super.getReferencedFields( metaData, element, query, parameter );
-    if ( referencedFields == null ) {
+    if ( referencedFields == null || referencedFields.length == 0 ) {
```
tmorgner: No. The sorting datafactory is a wrapper/decorator around other data factories and you just reverted the good change you made in the table-datafactory by redeclaring the fields as "don't know" instead of 'know for sure I have no fields'. Revert that.

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/cache/CachingDataFactory.java
ricardosilva88: Ok, just added because some assert on unit tests are failing due to this fact. Should I change test asserts instead?

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/sorting/SortingDataFactoryCore.java
ricardosilva88: Ok, just added because some assert on unit tests are failing due to this fact. Should I change test asserts instead?

PR #1271, opened 2019-05-23
[PRD-6054] Prevent caching the HTML attributes when they are obtained…

JoaoRMF:
… using expressions (they can change, and hence avoid caching them)


JoaoRMF:
@pentaho/kashyyyk and @tmorgner please review

tmorgner:
Yep, looks about right, except for that distinctive lack of test cases to validate the behaviour.

JoaoRMF:
> Yep, looks about right, except for that distinctive lack of test cases to validate the behaviour.

I'll add them... before we merge this!

Thanks

tmorgner:
Now it moved from good to perfect. Merged.

PR #1272, opened 2019-05-27
[PRD-6054] Prevent caching the HTML attributes when they are obtained…

JoaoRMF:
… using expressions (they can change, and hence avoid caching them)


JoaoRMF:
Master branch PR: https://github.com/pentaho/pentaho-reporting/pull/1271

JoaoRMF:
@kcruzada

PR #1304, opened 2019-11-26
[PRD-6087] - Getting IndexOutOfBoundsException when PageofPages in the Page Footer expand to the next line

bantonio82:
It was removed the property CollectSafeGuardingInformation from Thomas' commit, because with this property as **False** it would generate the same error as the one posted on PRD-6087 which would not give much information to the user, when **True** it will abort the processing and give to the user the reason for stopping the report processing.
So, if the reason for this case is to abort processing when the footer/header is change during the processing, I see no reason to keep the CollectSafeGuardingInformation and the related conditions.

@pentaho-lmartins @smmribeiro @ssamora

### DIFF
--- a/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/output/AbstractOutputProcessor.java
+++ b/engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/output/AbstractOutputProcessor.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2001 - 2013 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
+ * Copyright (c) 2001 - 2019 Object Refinery Ltd, Hitachi Vantara and Contributors..  All rights reserved.
  */
 
 package org.pentaho.reporting.engine.classic.core.layout.output;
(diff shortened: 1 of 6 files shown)


buildguy:
### :white_check_mark: Build finished in 11m 57s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl engine/core
```
---
:ok_hand: All tests passed!
 
Tests run: **2250**, Failures: **0**, Skipped: **15** &nbsp;&nbsp; [Test Results](http://wingman2.pentaho.net/job/projects/job/reporting/job/PR-1304/1/testReport)

 ---
 
 

:information_source:  _This is an automatic message_

engine/core/src/main/java/org/pentaho/reporting/engine/classic/core/layout/output/AbstractOutputProcessor.java
```diff
@@ -63,7 +98,11 @@ public final LogicalPageKey getLogicalPage( final int page ) {
       throw new IllegalStateException();
     }
 
-    return logicalPages.get( page );
+    try {
+      return logicalPages.get( page );
+    } catch ( IndexOutOfBoundsException e ) {
+      throw e;
+    }
```
smmribeiro: As there's no extra handling when an IndexOutOfBoundsException is caught, both old and new code behaves the same way: any exception thrown by `logicalPages.get( page )` will always be thrown back!
As so, the try-catch is dispensable and it's best to leave the code as it was.

buildguy:
### :white_check_mark: Build finished in 11m 43s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl engine/core
```
---
:ok_hand: All tests passed!
 
Tests run: **2250**, Failures: **0**, Skipped: **15** &nbsp;&nbsp; [Test Results](http://wingman2.pentaho.net/job/projects/job/reporting/job/PR-1304/2/testReport)

 ---
 
 

:information_source:  _This is an automatic message_

PR #1449, opened 2022-01-07
[CDF-946] PRD+CGG not working when there is a space in the parameter value

renato-s:
@andreramos89 
@bcostahitachivantara 
@smmribeiro

### DIFF
--- a/libraries/libloader/src/main/java/org/pentaho/reporting/libraries/resourceloader/loader/URLResourceLoader.java
+++ b/libraries/libloader/src/main/java/org/pentaho/reporting/libraries/resourceloader/loader/URLResourceLoader.java
@@ -12,7 +12,7 @@
 * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Lesser General Public License for more details.
 *
-* Copyright (c) 2006 - 2017 Hitachi Vantara and Contributors.  All rights reserved.
+* Copyright (c) 2006 - 2022 Hitachi Vantara and Contributors.  All rights reserved.
 */
 
 package org.pentaho.reporting.libraries.resourceloader.loader;
@@ -70,13 +70,16 @@ public class URLResourceLoader implements ResourceLoader {
       return new ResourceKey( SCHEMA_NAME, value, factoryKeys );
     }
     if ( value instanceof String ) {
-      final String valueString = (String) value;
+      String valueString = (String) value;
       // the 'file' schema is defined to have double-slashes, but the JDK
       // ignores RFC 1738 in that case. So we have to check for these beasts
       // directly ..
       if ( valueString.indexOf( "://" ) >= 0 || valueString.startsWith( "file:/" ) ) {
         try {
-          return new ResourceKey( SCHEMA_NAME, new URL( valueString ), factoryKeys );
+          // query parameters might contain ' ' (space) that must be escaped
+          // other 'special' characters shouldn't need escaping, as they are
+          // relaxedQueryChars in Tomcat config
+          return new ResourceKey( SCHEMA_NAME, new URL( valueString.replace( " ", "%20" ) ), factoryKeys );
 
         } catch ( MalformedURLException mfue ) {
           // we dont take this easy!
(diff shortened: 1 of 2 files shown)


libraries/libloader/src/main/java/org/pentaho/reporting/libraries/resourceloader/loader/URLResourceLoader.java
```diff
@@ -70,13 +70,22 @@ public ResourceKey createKey( final Object value, final Map factoryKeys ) throws
       return new ResourceKey( SCHEMA_NAME, value, factoryKeys );
     }
     if ( value instanceof String ) {
-      final String valueString = (String) value;
+      String valueString = (String) value;
+      final String encodedValue;
+      // query parameters might contain ' ' (space) that must be escaped
+      // other 'special' characters shouldn't need escaping, as they are
+      // relaxedQueryChars in Tomcat config
+      if ( valueString.contains( " " ) ) {
```
bcostahitachivantara, line 78: Why not replace this if else just by: encodedValue = valueString.replace( " ", "%20" );

If you do not have chars to replace, you do not do the replacement. In terms of performance, it is very similar because you do not have to call contains.

libraries/libloader/src/main/java/org/pentaho/reporting/libraries/resourceloader/loader/URLResourceLoader.java
```diff
@@ -70,13 +70,22 @@ public ResourceKey createKey( final Object value, final Map factoryKeys ) throws
...
+        encodedValue = valueString.replace( " ", "%20" );
+      } else {
+        encodedValue = valueString;
+      }
       // the 'file' schema is defined to have double-slashes, but the JDK
       // ignores RFC 1738 in that case. So we have to check for these beasts
       // directly ..
-      if ( valueString.indexOf( "://" ) >= 0 || valueString.startsWith( "file:/" ) ) {
+      if ( encodedValue.indexOf( "://" ) >= 0 || encodedValue.startsWith( "file:/" ) ) {
```
bcostahitachivantara, line 86: Can you ensure that when we are accessing to a file:// is spaces supported?

libraries/libloader/src/main/java/org/pentaho/reporting/libraries/resourceloader/loader/URLResourceLoader.java
```diff
@@ -70,13 +70,22 @@ public ResourceKey createKey( final Object value, final Map factoryKeys ) throws
...
         try {
-          return new ResourceKey( SCHEMA_NAME, new URL( valueString ), factoryKeys );
+          return new ResourceKey( SCHEMA_NAME, new URL( encodedValue ), factoryKeys );
```
bcostahitachivantara, line 88: Probably we could condense all changes in this line passing: valueString.replace( " ", "%20" ); to the new URL()

bcostahitachivantara (changes requested):
It is possible to write some unit test to coverage this changes?

bcostahitachivantara:
You could also change the logic to return null when the value does not contains //: or starts with file: this saves us the cost of replace spaces on string.

bcostahitachivantara:
Another important point, what happens if the string has more than one space?

buildguy:
### :white_check_mark: Build finished in 6m 31s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl libraries/libloader
```
---
 :exclamation: No tests found!
 ---
 
 

:information_source:  _This is an automatic message_

libraries/libloader/src/main/java/org/pentaho/reporting/libraries/resourceloader/loader/URLResourceLoader.java
renato-s: Yes, file:// supports the usage of spaces

libraries/libloader/src/main/java/org/pentaho/reporting/libraries/resourceloader/loader/URLResourceLoader.java
renato-s: Upon further investigation, it was found that, indeed, the contains provided no benefit and, as suggested, the validation was removed

libraries/libloader/src/main/java/org/pentaho/reporting/libraries/resourceloader/loader/URLResourceLoader.java
renato-s: After changes made based on the above suggestions, the lines were condensed to a single line to improve readability

bcostahitachivantara (approved):

buildguy:
### :white_check_mark: Build finished in 7m 53s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl libraries/libloader
```
---
 :exclamation: No tests found!
 ---
 
 

:information_source:  _This is an automatic message_

PR #1503, opened 2022-05-17
[SP-6077] Backport of PRD-6134 - Output to Excel does not create Excel footer if more than a certain number of records are on the sheet

andreramos89:
@smmribeiro 
@renato-s 
@bcostahitachivantara


smmribeiro (approved):
Consistent with the original PR.

PR #1597, opened 2023-10-19
[PDI-19936] - Formula Step : Unicode decimal code for surrogate pair characters is not returned correctly using CODE/UNICODE formula.

andreramos89:
Changing character reference to accommodate 4 bytes characters in Unicode

@smmribeiro
@renato-s
@bcostahitachivantara


renato-s (approved):

smmribeiro (approved):
Looks good!

